PR #4, opened 2025-02-16
add CI.yml

ufechner7:
add a workflow for running the tests automatically


1-Bart-1 (changes requested):
Looks good. Please remove the test_broken commit, as this was a mistake (the velocities are correct and the test should pass) and I changed the rtol to 1e-2 already.

ufechner7 (commented):
Fixed

PR #9, opened 2025-02-17
Add types MVec3, PosVector, VelVector

ufechner7:
- add package StaticArrays
- add the concrete type MVec3
- add the abstract types PosVector and VelVector
- use the abstract types in the function signatures for position and velocity vectors

Using MVec3 in structs will come in a second pull request.


src/VortexStepMethod.jl
```diff
@@ -18,6 +19,21 @@ export add_section!, set_va!
...
+"""
+const MVec3    = MVector{3, Float64}
+
+"""
+   const PosVector=Union{MVec3, Vector}
+
+Position vector, either a `MVec3` or a `Vector` for use in function signatures.
+"""
+const PosVector=Union{MVec3, Vector}
+const VelVector=Union{MVec3, Vector}
```
1-Bart-1, line 35: Why the different names for the same type?

ufechner7: Because a position vector is not the same as a velocity vector. You might want to have Float64 for the position and Float32 for the velocity, who knows?

But feel free to suggest a different, common name for this Union.

PR #50, opened 2025-02-22
Start documenting the structs and types

ufechner7:


src/filament.jl
```diff
@@ -12,12 +12,18 @@ const NU = 1.48e-5     # Kinematic viscosity of air
...
 """
 struct BoundFilament <: Filament
-    x1::Vector{Float64}    # First point
-    x2::Vector{Float64}    # Second point
-    length::Float64        # Filament length
-    r0::Vector{Float64}    # Vector from x1 to x2
+    x1::MVec3       # First point
+    x2::MVec3       # Second point
+    length::Float64 # Filament length
+    r0::MVec3       # Vector from x1 to x2
```
1-Bart-1, line 26: Why is the type of x1 and x2 inside the struct not the same as in the constructor?
And in general, we can make the code more flexible with:
```
struct BoundFilament{T} <: Filament
    x1::T       # First point
...
function BoundFilament(x1::T, x2::T) where T
        new{T}(x1, x2, norm(x2 - x1), x2 - x1)
```

ufechner7: This is on purpose. The signature of a function, and a constructor is a function should be flexible. The types in a struct should always be concrete for performance and compactness reasons.

By doing this calling the constructor with a Vector{Float64} will still work, but the data is stored in a compact and fast MVector.

PR #60, opened 2025-02-24
use SVec3 instead MVec3

ufechner7:


ufechner7:
Not compatible with last pull request.

PR #67, opened 2025-02-25
Remove interpolation allocation

1-Bart-1:


test/bench.jl
```diff
@@ -33,7 +33,7 @@ using LinearAlgebra
         [chord, -span/2, 0.0], # Right tip TE
         :inviscid)
     
-    body_aero = BodyAerodynamics([wing])
+    global body_aero = BodyAerodynamics([wing])
```
ufechner7, line 36: Why is this declared global?

examples/bench.jl
```diff
@@ -1,47 +1,42 @@
...
+using Pkg
 
-plot = true
-
-# Step 1: Define wing parameters
-n_panels = 20          # Number of panels
-span = 20.0            # Wing span [m]
-chord = 1.0            # Chord length [m]
-v_a = 20.0            # Magnitude of inflow velocity [m/s]
-density = 1.225        # Air density [kg/m³]
```
ufechner7, line 12: Why do you not need the air density in your new version?

ufechner7 (changes requested):
Apart from the one inline comment, I have the impression you replaced the old benchmark with a new benchmark. It would be better to keep the old benchmark and add the new benchmark, or to create second `bench2.jl` script for the new benchmark.

src/panel.jl
```diff
@@ -1,5 +1,8 @@
 using LinearAlgebra
 
+const I1 = Interpolations.FilledExtrapolation{Float64, 1, Interpolations.GriddedInterpolation{Float64, 1, Vector{Float64}, Interpolations.Gridded{Interpolations.Linear{Interpolations.Throw{Interpolations.OnGrid}}}, Tuple{Vector{Float64}}}, Interpolations.Gridded{Interpolations.Linear{Interpolations.Throw{Interpolations.OnGrid}}}, Float64}
```
ufechner7, line 3: What is that? Please add a comment for both I1 and I2.

ufechner7 (changes requested):
One more remark.

src/panel.jl
1-Bart-1: It was the solution for making the interpolations non-allocating. Had to be statically typed.

PR #70, opened 2025-02-26
use enum for parameter "model"

ufechner7:
First step to use enums.

I addition, improve doc strings.


1-Bart-1 (commented):
Will you change all Symbols to enums? Because this pull request causes a mix of symbols and enums in different places. The aero_model still uses a Symbol for example.

1-Bart-1 (approved):

1-Bart-1 (approved):

PR #71, opened 2025-02-26
add enum PanelDistribution

ufechner7:


1-Bart-1 (changes requested):
Why all caps, these are not abbreviations. I would either use the old names or CamelCase instead.

1-Bart-1 (approved):

PR #73, opened 2025-02-26
init! functions

### LINKED ISSUE #19: Dynamically change segments.
Change the angle of attack or trailing edge deformation of segments dynamically, with few allocations and without having to create an entire new Wing.

1-Bart-1:
Add init! functions to update wing_aero in a less allocating way. Step 1 for dynamically changing segments.


test/test_panel.jl
```diff
@@ -113,10 +116,10 @@ end
     end
 
     # Generate mock polar data
-    alphas = deg2rad.(-10:1:25)
+    alphas = deg2rad.(collect(-10:1:25))
     n_points = length(alphas)
-    polar_data = [zeros(n_points) for _ in 1:4]
-    big_polar_data = [zeros(n_points) for _ in 1:4]
+    polar_data = (zeros(n_points), zeros(n_points), zeros(n_points), zeros(n_points))
+    big_polar_data = (zeros(n_points), zeros(n_points), zeros(n_points), zeros(n_points))
```
ufechner7, line 122: Why do you try to avoid a loop here?

ufechner7 (approved):
Just one inline question

test/test_panel.jl
1-Bart-1: I use a tuple now, don't know how to loop this with a tuple.

PR #75, opened 2025-02-27
move ControlPlots to extension

ufechner7:
- this allows it to use the package without Python or any plotting library installed
- for running the examples, you need to do `using ControlPlots` before `using VortexStepMethod`. Furthermore, ControlPlots must be installed in the global environment.


README.md
```diff
@@ -18,10 +18,10 @@ if you haven't already. On Linux, make sure that Python3, Matplotlib and LaTeX a
 sudo apt install python3-matplotlib
 sudo apt install texlive-full texlive-fonts-extra cm-super
 ```
-Furthermore, the package `TestEnv` must be installed globally:
-```
-julia -e 'using Pkg; Pkg.add("TestEnv")'
+Furthermore, the packages `TestEnv` and `ControlPlots` must be installed globally:
 ```
+julia -e 'using Pkg; Pkg.add("TestEnv"); Pkg.add(ControlPlots)'
```
1-Bart-1, line 23: something went wrong with the code block here, not rendered correctly

1-Bart-1: Also, it is becoming a lot of work to run the examples. We should consider having a setup script that can do the setup for the user if they want.

README.md
```diff
@@ -62,6 +62,7 @@ Pkg.update()
 ```
 and then you can display a menu with the available examples:
 ```julia
+using ControlPlots
```
1-Bart-1, line 65: ControlPlots should be included in the examples. No-one will want to run the examples without plotting.

1-Bart-1 (changes requested):

README.md
ufechner7: The point is that you need to do `using ControlPlots` BEFORE `using VortexStepMethod`. And the function `menu()` is provided by VortexStepMethod. So adding a `using ControlPlots` to the examples would not help. Unless we discourage the use of the function `menu()` and encourage to use `include("examples/menu.jl")`.

PR #82, opened 2025-03-01
Refactor Section, use @with_kw; refactor Wing

ufechner7:
Goals:
- do not user inner constructors
- document both the structs and the outer constructor

Done:
- [x] Section
- [x] Wing
- [x] KiteWing

This makes testing and documenting easier because the structs and and the additional constructors can be tested and documented independently.


ufechner7:
The keyword parameters of the KiteWing constructor still need some explanation. @1-Bart-1 Can you provide that?

1-Bart-1 (approved):

PR #88, opened 2025-03-02
Add solve!(res::Results, ...)

ufechner7:
- [x] add solve_basic
- [x] add enum Status FEASIBLE INFEASIBLE FAILED
- [x] add struct Results
- [x] add solve!(res::Results, ... )
- [x] Rename Results to VSMSolution
- [x] add sol::VSMSolution as field to Solver and return the result from solve!


1-Bart-1 (changes requested):
I would prefer:
- Just one exported solve function: solve!
- Get rid of the dictionary and the solve() function, they are not necessary

ufechner7:
> * Just one exported solve function: solve!
> 
> * Get rid of the dictionary and the solve() function, they are not necessary

I don't agree. The dict is useful for the plotting function. A solve! function is always more complicated to use than a solve function and only justified in the context of integration in KiteModels.

Furthermore, for KiteModels we do not need all the distributions etc. They are only useful for kitedesign.

So I think it makes sense to have a small solve! function optimized for low allocations.

1-Bart-1:
It should be fine to use a struct in the plotting functions.
I don't think calculating the distributions is bad for performance, it is the allocations that hurt. So if we find a way to have a solve function with few allocations, it is no problem for performance that it calculates the distributions as well.

What I don't like is that you have multiple solve functions that give results in a different way. This makes it more difficult to learn how to use the package.

ufechner7:
> You can add the res struct as a field in the Solve struct, and let solve(solver) return the solver.res struct.

Good idea.

PR #92, opened 2025-03-03
Make the coordinate frames of the panels use the same convention as the kite body frame

1-Bart-1:
This way, the panel frames align with the body frame of a rectangular wing and they are easier to understand.

x_airf along chord from le to te
y_airf spanwise
z_airf = x_airf cross y_airf


jellepoland (approved):
All transformations have been made correctly
x -> z
y -> x
z -> y

PR #96, opened 2025-03-04
Add origin parameter to BodyAerodynamics constructor

1-Bart-1:
Wings are initialized in CAD frame and are translated to the kite body frame using the kite_body_origin keyword argument.


1-Bart-1:
Done

ufechner7 (approved):

PR #108, opened 2025-03-06
KiteWing should be centered at center of mass

1-Bart-1:
The center of mass field is removed, and the LE and TE interpolations are relative to the center of mass instead. So the [0,0,0] position is the center of mass.


ufechner7:
> relative to the COM instead.

What is COM ?

1-Bart-1:
> > relative to the COM instead.
> 
> What is COM ?

Center of Mass

PR #115, opened 2025-03-09
Plot polars

1-Bart-1:
Move polar matrix plotting functions to the plotting extension and add Xfoil and Timers as normal dependencies


test/data/ram_air_kite_foil.dat
```diff
@@ -0,0 +1,45 @@
+ 1.0000000     0.0000479
```
ufechner7, line 1: Why are you adding this dat file? Is it input data? If it is input data, where is the data format and its usage it documented?

ufechner7 (commented):
Please, address my remarks.

scripts/polars.jl
```diff
@@ -133,42 +132,6 @@ try
         return lower_trailing_edge, upper_trailing_edge
     end
 
-    function replace_nan!(matrix)
```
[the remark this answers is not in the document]
1-Bart-1: It was not used here

test/data/ram_air_kite_foil.dat
1-Bart-1: Just used for testing. Only used in the test_plots for now, but can also be used later on for accuracy testing of the ram air kite. Added the bin files in the same directory so they don't have to be generated every time the package is tested.

examples/ram_air_kite.jl
```diff
@@ -1,14 +1,6 @@
 using ControlPlots
 using VortexStepMethod
 using LinearAlgebra
-using Pkg
```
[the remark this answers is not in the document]
1-Bart-1: CSV and DataFrames were not used in this example.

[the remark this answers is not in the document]
ufechner7: OK

scripts/polars.jl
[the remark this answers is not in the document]
ufechner7: OK

test/data/ram_air_kite_foil.dat
ufechner7: Added the bin files in the same directory so they don't have to be generated every time the package is tested.

Doesn't that create conflicts when they are re-generated?

1-Bart-1: They are not re-generated if they exist already

ufechner7: OK

PR #130, opened 2025-03-20
Refactoring, one bugfix

ufechner7:
- [x] fix "calculate_circulation_distribution_elliptical_wing()" never called
- [x] reduce allocations of the function mentioned above
- [x] add enum WingType and document it
- [x] use this enum instead of symbols in the code
- [x] use == instead of === to compare enum members
- [x] fix tests
- [x] update CI
- [x] address remarks
- [x] rebase and fix conflicts


ufechner7:
```
julia> include("test/bench2.jl")
Rectangular wing, solve_base!:
  493.861 μs (51 allocations: 12.61 KiB)
Rectangular wing, solve!:
  555.817 μs (85 allocations: 14.92 KiB)
```
This looks fine. But we have no test that proves that **calculate_circulation_distribution_elliptical_wing()** is called if needed.

mwes/mwe_03.jl
```diff
@@ -0,0 +1,47 @@
+# fix allocations of this array comprehension:
```
1-Bart-1, line 1: Add better comment explaining what this mwe does

src/VortexStepMethod.jl
```diff
@@ -242,4 +243,45 @@ include("body_aerodynamics.jl")
 include("wake.jl")
 include("solver.jl")
 
+@setup_workload begin
```
1-Bart-1, line 246: I prefer to have this in a seperate file "src/precompile.jl"

src/solver.jl
```diff
@@ -385,7 +385,7 @@ function solve_base!(solver::Solver, body_aero::BodyAerodynamics, gamma_distribu
     # Initialize gamma distribution
     gamma_initial = cache_base[1][solver.sol.chord_array]
     if isnothing(gamma_distribution)
-        if solver.type_initial_gamma_distribution === :elliptic
+        if solver.type_initial_gamma_distribution == ELLIPTIC
```
1-Bart-1, line 388: should be a triple equal sign `solver.type_initial_gamma_distribution === ELLIPTIC`

1-Bart-1 (changes requested):

src/solver.jl
ufechner7: No, to compare integers (and enum members are integer constants) == is prefered.

mwes/mwe_03.jl
ufechner7: I improved mwe_03.jl to be a better tutorial example.

src/VortexStepMethod.jl
ufechner7: Done.

1-Bart-1 (approved):

PR #138, opened 2025-03-24
Add job check_versions

ufechner7:
If there is no pre-release, don't runs the tests twice with the same Julia version (1.11).


ufechner7:
Too much work.

PR #152, opened 2025-03-28
Add bench_solve.jl

ufechner7:
Add a benchmark to track the performance of the solve! on one and the same PC, not part of the CI testing.


test/bench_solve.jl
```diff
@@ -0,0 +1,57 @@
+# DO NOT DELETE, DO NOT MODIFY!
+# This script will not be included in run_tests.jl
+# It shall only be executed on one and the same machine for benchmarking the solve functions
```
1-Bart-1, line 3: Can you specify the machine you are using? It says Ryzen 950x later in the code, but you can write it here as well, and specify how much RAM / cores you have.

PR #203, opened 2025-10-26
Unify Wing and RamAirWing

1-Bart-1:
This makes it possible to use the RamAirWing features (like reading from .obj file) for the Wing as well, while keeping backwards compatibility.


1-Bart-1:
replaced by #204

PR #227, opened 2026-04-23
Fix bug in untested linearize branch

1-Bart-1:


ufechner7 (approved):

ufechner7:
Perhaps wait for the comments from Copilot and add a description.

PR #232, opened 2026-05-05
Add forwarddiff compatibility and use it for linearisation

1-Bart-1:


examples/linearize_check.jl
```diff
@@ -0,0 +1,214 @@
+using Pkg
```
ufechner7, line 1: Can explain the purpose of the example? What does it show?

src/body_aerodynamics.jl
```diff
@@ -238,13 +238,13 @@ Initialize a BodyAerodynamics struct in-place by setting up panels and coefficie
 # Returns
 nothing
 """
-function reinit!(body_aero::BodyAerodynamics;
+function reinit!(body_aero::BodyAerodynamics{P, W, T};
```
ufechner7, line 241: This looks like a breaking change. Correct?

ufechner7 (commented):
Pretty large change. Did you test if it has a significant advantage?

1-Bart-1:
Significant advantage: linearization is more correct. Noise in the solutions significantly affected the quality of the jacobian. You can run examples/linearize_check.jl to see this.

src/body_aerodynamics.jl
1-Bart-1: No not breaking

examples/linearize_check.jl
1-Bart-1: see comment below

ufechner7 (approved):

PR #240, opened 2026-05-23
Use smooth norm and sqrt

1-Bart-1:


ufechner7 (commented):
Where is smooth_norm3 defined?
And why do you want to use it in the first place?

1-Bart-1:
https://github.com/OpenSourceAWE/VortexStepMethod.jl/pull/240/changes#diff-7c00982760bc184a65a15a5fd0a17038075d481dd7fe413f60100eed1a302861R325

defined here. And I use it for auto diff differentiability at zero.

ufechner7:
I would be happier with this PR if you could provide an example that fails without this PR. I am somewhat afraid that it hides errors.

1-Bart-1:
not necessary

PR #247, opened 2026-06-27
Add Li/Gaunaa spanwise artificial viscosity for post-stall stabilization

1-Bart-1:
Ports the opt-in implicit artificial-viscosity regularization (Li, Gaunaa, Pirrung & Lønbæk, TORQUE 2026) from the Python [Vortex-Step-Method](https://github.com/awegroup/Vortex-Step-Method/commit/2a12f13388122eafa541041c11c13be5f020f1fb) into the Julia LOOP solver.

## Why
Post-stall (negative lift-slope) circulation distributions develop non-physical sawtooth oscillations in the fixed-point `gamma_loop!` and never converge. The implicit scheme `(I - diag(mu) L) gamma = F(gamma)` has the same steady solution as the explicit fixed point but stays stable at relaxation factors of order one, whereas the explicit stable step shrinks like `N^-2` in post-stall.

## What
- **`Solver`**: new opt-in fields `is_with_artificial_viscosity` (default `false`) and `artificial_viscosity_factor` (default `0.035`), threaded through the `VSMSettings` and ForwardDiff dual-copy constructors.
- **`build_spanwise_laplacian!`**: discrete spanwise Laplacian with the second-order tip closures (Eq. 15) enforcing `gamma -> 0` at the tips.
- **`local_lift_slope!`**: per-panel `dCl/dalpha` via central difference of `calculate_cl`.
- **`gamma_loop!`** (LOOP): builds the Laplacian and planform area once, then when any panel is post-stall, solves `(I - diag(mu) L) gamma = F(gamma)` with `mu_i = max(0, -k S Cl'_i / dz_i^2)` before relaxation.
- **`SolverSettings`**: the two fields exposed for YAML config.

## Design note (difference from Python)
The Python version gates the expensive linear solve on per-panel stall angles read from raw polar arrays. In Julia the polar is an interpolation object, so the solve is instead gated on `any(mu .> 0)` — behaviourally identical (it fires exactly in post-stall) without exposing internal polar storage.

The feature is fully opt-in and a no-op in attached flow, so existing results are unchanged.

## Tests
Ports both Python tests:
- Laplacian tip closures (interior stencil + Eq. 15 closures).
- Post-stall stabilization on the paper's canonical AR=10 rectangular wing (plain iteration diverges; implicit scheme converges to a smooth distribution).

Verified locally: solver tests 14/14, ForwardDiff tests 7/7, and an integration smoke test confirming attached flow is bit-identical to baseline while a high-AoA case converges.


1-Bart-1:
Rebased onto current `main` (resolved the one `docs/src/private_functions.md` list conflict). Validated on Julia 1.12: Spanwise Laplacian tip closures 5/5, post-stall stabilization 3/3, ForwardDiff 7/7, solver constructor 3/3 — all green against main's split-`calc_forces!`/split-cd-cm refactors.

PR #262, opened 2026-08-17
Continuous wrapping method

1-Bart-1:
Split out of #258, which carried this rewrite alongside unrelated section-order work and failed CI because of it.

## What changes

`shrink_wrap` was a distance field: threshold at the ball radius, flood fill, erode back to `clearance`, trace the level set with marching squares, fair the result. Every step was bounded by the grid, so `cell_size` set the fidelity of the wrap and padding the field for `min_concave_radius` cost time in both directions.

`pivot_contour` now rolls the ball around the cloud and emits what its contact side sweeps: an arc of radius `clearance` about each point it touches, joined by an arc of the ball radius across each gap it cannot enter. Those arcs are the wrap boundary, so there is no resolution to choose.

- `ShrinkWrap`'s `cell_size` becomes `min_clearance` — a floor on `clearance` rather than a grid pitch. Still accepted under the old name.
- The wrap sits at exactly `clearance` from the cloud instead of a cell over it: a V3 canopy's aft strip is `2 * clearance` thick where the grid gave `3.5 * cell_size`.
- `min_concave_radius` stops costing anything. One V3 slice at radius 0.4 drops from 173 ms to 10 ms; at the default radius, 15 ms to 3 ms.

## Resampling had to follow

`push_arc!` steps each arc to a chord tolerance, so its segments scale as `sqrt(radius)` and the trace's node spacing alternates by a factor of several between the two arc families. Diffusing the turning *per node* carried that alternation into the sampling measure, leaving resampled panel lengths oscillating by up to 5.7x between neighbours.

- `smoothed_curvature` averages the turning over a fixed *arclength* band instead, which is independent of how the arcs happen to be stepped. Neighbour ratio drops to the 3.0 of a plain cosine-clustered `.dat`.
- `enforce_min_spacing!` keeps cosine clustering — which thins as the square of the station count — from putting leading-edge panels a fraction of a percent of chord long.

## Known limitation — not ready to merge

XFoil's viscous solver still fails to converge on the wrapped NACA0012 in `test/solver/test_backend_comparison.jl`, where the distance-field wrap succeeded, so that test errors in `interpolate_matrix_nans!` with nothing to interpolate from.

What is established so far:

- XFoil reports no `LEFIND` warning, so the contour is well formed; it is the viscous march that fails, returning an alpha-independent cl (~-0.484 at -4, 0 and +4 deg alike).
- `repanel=true` converges 6/6, so XFoil's own repaneling repairs whatever it dislikes.
- Convergence is **non-monotone** in the panel size floor: 0.1 gives 6/6, 0.2-0.5 gives 0/6, 0.7 gives 6/6. So neither grading nor absolute panel size is the discriminator, and tuning a constant until it passes would not be a fix.
- Current suspect: the round trailing-edge cap of radius `clearance`. The contour both starts and ends inside it, so each surface turns ~90 deg around the cap before the boundary layer can start. A sharp-TE airfoil has no such cap in the distance-field wrap.

Next step is to confirm that mechanism and, per the validation bar this repo needs, check any fix through a written `.dat` roundtrip plus ~1e-7 coordinate noise over several sections rather than a single in-memory pass.


1-Bort-1:
### Fixed

Rebased onto `main`. Confirmed the round-trailing-edge-cap mechanism: `shrink_wrap`
floored `clearance` at `min_clearance` for every convex corner, so a closed airfoil
wrapped at `clearance=0` grew a round TE cap of radius 0.001. XFoil flags that contour
`sharp` (first point == last) while the geometry is blunt, so the viscous march fails
and the blunt TE shifts the Kutta point (cl ±0.486 vs ±0.434 raw).

### Changes

- `src/airfoil_aero/shrink_wrap.jl:338` — a closed loop (first ≈ last cloud point) now
  uses its true `clearance` instead of `max(clearance, min_clearance)`, so
  `clearance=0` hugs the input and keeps the sharp TE sharp; an open single-membrane
  cloud still gets the `min_clearance` capsule. The base NACA0012 now converges 6/6 in
  `test_backend_comparison.jl` with cl ±0.432.
- `test/airfoil_aero/test_airfoil_aero.jl:73` — regression test that the TE stays sharp
  (`maximum(xt) ≈ 1.0`) at zero clearance.

### Verification

- [x] `test/airfoil_aero/test_airfoil_aero.jl` green on the box (exit 0)
- [ ] `test/solver/test_backend_comparison.jl` — **14/15**: the `delta=1` column still
      returns NaN at alpha=-4 (borderline at +6), so line 29 still fails.
- Risk: this is only the TE half of the regression. The deflected -4/+6 residual is a
  *resampling* sensitivity (not the cap): the raw deflected NACA0012 converges 6/6, and
  `n_points=80` (159 pts) does too, but the 239-pt default diverges; it is non-monotone
  in gap (0→4/6, 1e-4→6/6, 5e-4→0/6), `curvature_weight` and `min_concave_radius`, so
  tuning a constant will not be robust.

### Commits

- `4a926cb` Keep refined surface tables through a coarsening remesh
- `912c86b` Decide panel spanwise orientation once per wing
- `0d2120f` Drop the BREAKING note on the panel reinit! signature
- `aebfd96` Optional flow-curvature pitch-rate moment
- `2ac530b` Normalize wing section order to +y -> -y on load
- `5f27ade` Different field for corrected vs uncorrected
- `9364a2b` Update default manifests
- `89641dd` Document the new span-order and table-path helpers
- `3340832` Non-interactive Julia selection for bin/install
- `0621b27` Backtick the underscored @ref targets
- `2030908` Leave the shrink-wrap rewrite out of this branch
- `3d627b1` Leave the Section method of reinit! undocumented
- `8c0a647` Validate the moment relation against measured section data
- `567c129` Merge pull request #258 from OpenSourceAWE/fix/refined-surface-tables
- `60c2651` Merge pull request #260 from OpenSourceAWE/feat/flow-curvature
- `179eca9` Bump version
- `34f43c3` Rotate a twisted section's chord fully about the spanwise axis
- `05166ca` Merge pull request #263 from OpenSourceAWE/fix/twist-rodrigues-axial-term
- `eea12cf` Take the billow axis from the section pair, not a global projection
- `c88bb46` Merge pull request #264 from OpenSourceAWE/fix/billow-axis-span-order
- `a385be7` Bump version
- `eee5982` Write the panel aerodynamics once, as traceable functions
- `1a4da79` Update version
- `dd97f12` Reconstruct NeuralFoil's surface pressure from edge velocity in arc length
- `19817fd` Regenerate polars from the deformed shape instead of a flap angle
- `d62f83e` Continue a TAYLOR polar linearly past its fit window
- `34987f8` Say what a live polar refresh actually allocates
- `dec31c9` Draw the airfoil the live polars were actually generated from
- `74fb314` Put panel_contour ahead of the skin's docstring
- `5c42181` Take the chord line out of a deflection, and never return a negative drag
- `db488d7` Keep generated aero tables out of git
- `7c304b1` Sample a live polar instead of fitting one
- `27e2b1a` Put a live polar next to XFoil on the shape it was generated from
- `0793622` March XFoil onto the deformed shape instead of jumping to its angle
- `ba28226` Give chord_line its own docstring and file the AirfoilAero privates
- `5a2e796` Backtick the identifier cross-references Documenter 1.18 now requires
- `89af20b` A live polar is a polar, not a model of its own
- `dcebe32` Let a short polar table scan for its knot instead of bisecting
- `fbdbaef` Take a live polar onto a wing whose table carries a delta axis
- `84d3869` Add polar_model to the manual
- `9249f98` Bump version
- `92987eb` Let a polar refresh write into the panel instead of rebuilding it
- `becd3ba` Let a live polar refresh run without touching the heap
- `67dc29b` Make KulfanParameters mutable so a panel holds the shape, not a copy
- `6fadaf9` Note what the allocation work buys a multi-core sweep
- `394859e` Measure the refresh allocations without boxing its return value
- `fa4dc89` Bump version
- `ca2d430` Trace the shrink wrap's rolling ball exactly instead of on a grid
- `523d31a` Describe the rolling-ball wrap where the distance field was documented
- `8c10677` Keep a closed loop's trailing edge sharp at zero clearance

[Review just these changes](https://github.com/OpenSourceAWE/VortexStepMethod.jl/pull/262/files/db00f137696ed5ecefe70d3cc1d29207ba659c7f..8c10677acd71232254726b74bf3c568aa7fc17b7)

Task `VortexStepMethod.jl-262`

1-Bort-1:
**Local full suite: FAIL** (0 min, Julia 1.12.7, one cell of the matrix)

```
  [7] #invokelatest_gr#239
    @ ./reflection.jl:1297 [inlined]
  [8] invokelatest_gr
    @ ./reflection.jl:1289 [inlined]
  [9] maybe_cachefile_lock(f::Base.var"#__require_prelocked##0#__require_prelocked##1"{Base.PkgId, String, Dict{String, Int64}, Base.RefValue{Bool}}, pkg::Base.PkgId, srcpath::String; stale_age::Int64)
    @ Base ./loading.jl:3964
 [10] maybe_cachefile_lock
    @ ./loading.jl:3961 [inlined]
 [11] __require_prelocked(pkg::Base.PkgId, env::String)
    @ Base ./loading.jl:2735
 [12] _require_prelocked(uuidkey::Base.PkgId, env::String)
    @ Base ./loading.jl:2560
 [13] macro expansion
    @ ./loading.jl:2488 [inlined]
 [14] macro expansion
    @ ./lock.jl:376 [inlined]
 [15] __require(into::Module, mod::Symbol)
    @ Base ./loading.jl:2453
 [16] require(into::Module, mod::Symbol)
    @ Base ./loading.jl:2429
 [17] top-level scope
    @ ~/.julia/packages/MakieControlPlots/hXwA9/src/MakieControlPlots.jl:3
 [18] include(mod::Module, _path::String)
    @ Base ./Base.jl:306
 [19] include_package_for_output(pkg::Base.PkgId, input::String, depot_path::Vector{String}, dl_load_path::Vector{String}, load_path::Vector{String}, concrete_deps::Vector{Pair{Base.PkgId, UInt128}}, source::Nothing)
    @ Base ./loading.jl:3106
 [20] top-level scope
    @ stdin:5
 [21] eval(m::Module, e::Any)
    @ Core ./boot.jl:489
 [22] include_string(mapexpr::typeof(identity), mod::Module, code::String, filename::String)
    @ Base ./loading.jl:2952
 [23] include_string
    @ ./loading.jl:2962 [inlined]
 [24] exec_options(opts::Base.JLOptions)
    @ Base ./client.jl:315
 [25] _start()
    @ Base ./client.jl:550
in expression starting at /home/agent/.julia/packages/MakieControlPlots/hXwA9/src/MakieControlPlots.jl:1
in expression starting at stdin:5
```

src/airfoil_aero/shrink_wrap.jl
```diff
@@ -303,85 +320,26 @@ end
...
-        d = dist_at(cx, cy)
-        d >= gap && return (cx, cy)
-        h = cell / 2
-        gx = (dist_at(cx + h, cy) - dist_at(cx - h, cy)) / cell
-        gy = (dist_at(cx, cy + h) - dist_at(cx, cy - h)) / cell
-        len = hypot(gx, gy)
-        len < 1e-9 && return (cx, cy)
-        return (cx + (gap - d) * gx / len, cy + (gap - d) * gy / len)
-    end
+    px, py = pivot_contour(xn, yn, ball, gap)
```
1-Bort-1: MAJOR: The PR card itself says the new `pivot_contour` wrap still errors XFoil's viscous march in `test/solver/test_backend_comparison.jl` ('not ready to merge'), and the diff adds no tests; wiring it in here as the only wrap path keeps the suite red.

src/airfoil_aero/shrink_wrap.jl
```diff
@@ -243,54 +207,107 @@ function largest_linking_gap(x, y)
...
 ends attract points and sharp features are resolved by several panels regardless of
-their size (like XFoil's curvature-attracted paneling). Endpoints are preserved.
+their size (like XFoil's curvature-attracted paneling). The turning is taken as a
+curvature density smoothed over a few panel lengths ([`smoothed_curvature`](@ref)), so
+the measure grows smoothly and neighbouring panels stay comparable in length. No panel
+is left shorter than `min_panel_fraction` of the mean
+([`enforce_min_spacing!`](@ref)). Endpoints are preserved.
 """
-function resample_arc(ax, ay, n, curvature_weight)
+function resample_arc(ax, ay, n, curvature_weight; min_panel_fraction=0.25)
```
1-Bort-1: MINOR: `min_panel_fraction=0.25` hardcodes the very knob the card swept (0.1 converges, 0.2-0.5 fail, 0.7 converges) and lands inside its failing band; since the card concludes tuning this constant 'would not be a fix', the default here is unprincipled.

src/airfoil_aero/shrink_wrap.jl
```diff
@@ -243,54 +207,107 @@ function largest_linking_gap(x, y)
...
-few panels to near-zero length, which XFoil's viscous solver cannot handle. Spreading
-the turn over a short band refines a group of panels gradually instead, the way
-XFoil's `PANGEN` bunches panels on a *smoothed* curvature. The total turn is
-conserved, so leading-edge clustering is preserved.
+Turning per unit arclength at each node, averaged over a window of length `band`
+centred on it. Averaging over a length rather than over a node count is what makes it
+independent of how the traced arcs happen to be stepped: [`push_arc!`](@ref) steps each
+arc to a chord tolerance, so its segments scale as `sqrt(radius)` and the trace's node
+spacing alternates by a factor of several between the `clearance` arcs and the wider
+pivot arcs. Diffusing per node would carry that alternation into the measure and leave
```
1-Bort-1: MINOR: This docstring argues a rejected alternative ('Diffusing per node would carry that alternation into the measure'), which the what-not-why rule forbids; the reasoning belongs in the PR's Decisions section, not frozen in code.

src/airfoil_aero/shrink_wrap.jl
```diff
@@ -1,212 +1,176 @@
...
-    nx, ny = size(field)
-    return function (px, py)
-        gi = clamp((px - x0) / cell + 1, 1.0, nx - 1e-9)
-        gj = clamp((py - y0) / cell + 1, 1.0, ny - 1e-9)
-        i, j = floor(Int, gi), floor(Int, gj)
-        ti, tj = gi - i, gj - j
-        return (field[i, j] * (1 - ti) + field[i+1, j] * ti) * (1 - tj) +
-               (field[i, j+1] * (1 - ti) + field[i+1, j+1] * ti) * tj
+function pivot_contour(x, y, r, clearance)
+    start = argmin(x)
```
1-Bort-1: MINOR: Pivoting from `argmin(x)` silently ignores disconnected cloud parts; the deleted code warned when a comparable secondary contour was discarded. Multi-cluster slices now wrap one cluster with no signal.

1-Bort-1 (commented):
<sub>Superseded by the review further down this thread.</sub>

src/airfoil_aero/shrink_wrap.jl
```diff
@@ -303,85 +327,29 @@ end
...
-        d >= gap && return (cx, cy)
-        h = cell / 2
-        gx = (dist_at(cx + h, cy) - dist_at(cx - h, cy)) / cell
-        gy = (dist_at(cx, cy + h) - dist_at(cx, cy - h)) / cell
-        len = hypot(gx, gy)
-        len < 1e-9 && return (cx, cy)
-        return (cx + (gap - d) * gx / len, cy + (gap - d) * gy / len)
-    end
+    ball = max(method.min_concave_radius, 1.01 * linking / 2)
+    px, py = pivot_contour(xn, yn, ball, gap)
```
1-Bort-1, line 349: MAJOR: The PR card itself declares this 'Known limitation — not ready to merge': XFoil's viscous solver still fails on the wrapped NACA0012 (exercised via deform_section here and in test/solver/test_backend_comparison.jl), which the round clearance-radius trailing-edge cap now produces. Shipping a known-failing validation test, before the .dat-roundtrip-plus-noise check the author lists as the repo's bar, is blocking.

1-Bort-1 (commented):

**Verdict: REQUEST CHANGES** · 1 inline, 0 off the diff

**Good**

- The entire distance-field subsystem (distance_parabolas!, squared_distance_transform!, grid_sampler, flood_outside, trace_level_set, smooth_turning!) is deleted cleanly with no dangling callers — grep confirms zero remaining references, satisfying the §2 removal rule.
- cell_size → min_clearance is a documented, backward-compatible migration: the constructor keeps a cell_size=nothing alias, so existing ShrinkWrap(cell_size=...) call sites keep working.
- Docs stay in sync: airfoil_pipeline.md prose and private_functions.md's @docs list both reflect the new rolling-ball names with the old entries removed.
- The resampling change (smoothed_curvature over an arclength band, enforce_min_spacing! floor) is well documented and directly targets the two defects named in the PR card (per-node diffusion and cosine thinning).

**Not good**

- `src/airfoil_aero/shrink_wrap.jl:349` — The PR card itself declares this 'Known limitation — not ready to merge': XFoil's viscous solver still fails on the wrapped NACA0012 (exercised via deform_section here and in test/solver/test_backend_comparison.jl), which the round clearance-radius trailing-edge cap now produces. Shipping a known-failing validation test, before the .dat-roundtrip-plus-noise check the author lists as the repo's bar, is blocking.
- The outer pivot bound 4*length(x)+100 (and the +100) is an unexplained magic cap with no justification in docstring or PR.
- cell_size is accepted as a silent alias with no deprecation path, so the old name can linger forever — §3 discourages renaming API without a removal plan.
- private_functions.md lists enforce_min_spacing! before resample_arc and smoothed_curvature last, out of source order, which is only cosmetic.

Issue #287, opened 2026-09-09
test_forwarddiff.jl's POLAR_MATRICES check flakes at ~4%, and the oper


1-Bort-1:
test_forwarddiff.jl's POLAR_MATRICES check flakes at ~4%, and the operating point 320e756 chose to prevent it does not move

`test/solver/test_forwarddiff.jl:88` asserts `rel_err < 1e-3` on `max|jac_fwd - jac_fd| / max|jac_fwd|`. It failed on #285's CI in `Julia 1.12 - ubuntu - x64 - with code coverage` at **0.0400**, with the other five matrix cells green on the same commit. #285's diff is entirely inside the `NONLIN` branch of `gamma_loop!` (`src/solver.jl:913-1010`, which `return`s before the `LOOP` branch at `:1012`) and this test builds a `solver_type=LOOP` solver, so no changed line is reachable from it.

This is the third time the check has been in the frame. 320e756 recorded "one Windows run crossed 1e-3 at **0.044** while a re-run passed", moved `y_op[1:4]` off zero (`defl_rad = deg2rad(1.0)`) and loosened the bound to 0.1 in the same commit. ae11d48 then pinned NeuralFoil table generation to one BLAS thread, tightened the bound back to 1e-3, and put `y_op[1:4]` back to `zeros(4)`. 0.0400 is the same jump as 0.044.

**Measured here** (VSM at the test's operating point, 8 panels, tables regenerated on this box):

- `rel_err = 3.31e-6`, repeatable to four significant figures — 300x inside the 1e-3 bound. CI's 0.0400 is 12000x that, so it is a discrete derivative jump, not tolerance drift.
- Flat in freestream alpha: 3.23e-6 at 5.0 deg, 3.23e-6 at 6.0 deg, 3.31e-6 at 7.5 deg, 3.23e-6 at 10.0 deg.
- **Flat in theta**: 3.314e-6 at theta = 0, and the same 3.314e-6 at 1e-12, -1e-12, 1e-9, -1e-9, 1e-6 and -1e-6; 3.316e-6 at 0.25 deg, 3.273e-6 at 1.0 deg, 3.307e-6 at -1.0 deg. There is no knife edge in theta anywhere between 0 and 1 deg.

[...]

Issue #279, opened 2026-09-07
cleanup: CI.yml carries `fail-fast: false`, so a broken matrix runs every job to the end


1-Bort-1:
Noticed while confirming main was green for the v4.4.0 release (#276). Not
blocking that release — a `cleanup:` PR.

`.github/workflows/CI.yml:22` and `:78` both set `fail-fast: false`. AGENTS.md §4
step 10 asks the matrix to stop at the first failing job, so an answer comes back
sooner and the remaining runners are left for other work; the box's local mirror
is already set that way.

The tradeoff is worth stating rather than assuming: with `fail-fast: true` a
failure on one cell cancels the others, so you lose the "is this Windows-only or
everywhere?" signal in the same run and have to re-run to get it. On a matrix
this wide (Julia 1.11/1.12 × ubuntu/windows/macOS, plus coverage jobs) that
signal is worth something. Whoever picks this up should decide which of the two
costs more here.

1-Bort-1:
Plan first, as asked: removing both `fail-fast: false` lines from .github/workflows/CI.yml (AGENTS.md §4.10), stacked on #289 which is open on the same two job blocks. Gathering the CI-history numbers that decide the tradeoff you named; no code until the plan is approved.

1-Bort-1:
# Review plan — #279, `fail-fast: false` in CI.yml

## Objective

Settle, with numbers from this repo rather than assumption, whether CI.yml's two
`fail-fast: false` lines should go — which is what the issue asked for. AGENTS.md
§4 step 10 says they should. Measured on the last 40 CI runs, they should not, so
the objective of this turn is the decision, not the edit.

## Approach

Measured the window 2026-09-01 → 2026-09-11: 40 runs, 280 jobs, 5263
runner-minutes. Three numbers decide it.

- **9 of 9 failing runs failed in exactly one matrix cell; 0 of 9 failed in all
  of them.** Every failure was on Julia 1.12 with the 1.11 sibling green in the
  same run. That green sibling is what fail-fast cancels.
- **Fail-fast would have reclaimed 51.9 runner-minutes over those 9 runs — 0.99%
  of the window** — with a median of 2.1 minutes per failing run and two runs
  saving nothing. Failures land at a median 79% of a passing job's wall clock,
  so the cheap cells have already finished by then. Queue wait is a median 2 s
  (p90 152 s), and the repo is public, so nothing is waiting for the capacity.
[...]

PR #289, opened 2026-09-09
CI: the code-coverage job's timeout-minutes is 30, tighter than the pl

### LINKED ISSUE #288: CI: the code-coverage job's timeout-minutes is 30, tighter than the pl
CI: the code-coverage job's timeout-minutes is 30, tighter than the plain test job's 40, and it is the slower of the two

`.github/workflows/CI.yml:76` gives `test-with-code-coverage` `timeout-minutes: 30`; `.github/workflows/CI.yml:20` gives the plain `test` job 40. The coverage job runs the same suite under coverage instrumentation and then adds `julia-processcoverage` and a Codecov upload, so it is the one that needs the longer cap, not the shorter.

Seen on #285, same commit, two attempts of `Julia 1.12 - ubuntu-latest - x64 - with code coverage`:

- attempt 1: job 16m5s, `julia-runtest` 15m24s — inside the cap, and it reported a test failure (that one is #287).
- attempt 2: `julia-runtest` green (`Testing VortexStepMethod... | 6304 passed, 1 broken, 0 failed, 10m06s`), every step ticked through `Complete job` — and the job was cancelled at **30m1s** with `The job has exceeded the maximum execution time of 30m0s`.

So the cap sits at roughly twice a warm run, and a cold `julia-actions/cache@v2` restore tips it over. The failure it produces is indistinguishable at a glance from a real one: the check goes red with the suite entirely green.

For comparison, `Julia 1.12 - ubuntu-latest - x64 - pull_request` took 28m54s on the same commit under its 40-minute cap.

What I would do: raise `test-with-code-coverage` to at least 40 to match the plain job, or 50 given it is doing strictly more work. Separate from #279, which is about `fail-fast: false` in the same file.

1-Bort-1:
### TL;DR

`test-with-code-coverage` had `timeout-minutes: 30` while the plain `test` job, which does strictly less work, had 40 — so the slower job was the one that got killed. Raised the coverage job to 50.

### What the cap was actually cutting off

The job that was cancelled on #285 had finished. Attempt 2 of run [34401975496](https://github.com/OpenSourceAWE/VortexStepMethod.jl/actions/runs/34401975496/attempts/2), `Julia 1.12 - ubuntu-latest - x64 - with code coverage`: every step reports `success`, `Complete job` included, and the job is `cancelled` at 30m01s with `The job has exceeded the maximum execution time of 30m0s`.

The steps of that attempt: setup, cache and buildpkg 1m02s · `julia-runtest` 28m15s · `julia-processcoverage` 0m24s · `codecov` 0m02s · post-steps 0m15s. That is ≈29m58s of work against a 30m00s cap. Inside the 28m15s of `julia-runtest`, the suite itself was 10m06s (`6304 passed, 1 broken, 0 failed`); the other ~18 minutes were the cold precompile of the test environment after a `julia-actions/cache@v2` miss.

So this is not one unlucky run. Over the last 11 CI runs the coverage jobs land at 9, 10, 10, 10, 10, 12, 14, 14, 14, 16, 16, 17, 20, 20, 21, 26, 27, 28, 28 and 28 minutes — warm around 10–16, cold around 26–28. Three green runs came within two minutes of the cap. The plain `test` ubuntu jobs over the same runs take 21–29 minutes under their 40-minute cap, so the job doing less work had the more generous margin.

I went with 50 rather than 40 because 40 would only restore parity with the plain job, and the coverage job adds instrumentation, `julia-processcoverage` and the Codecov upload on top of the same suite. 50 against a 28-minute cold run is roughly the margin `test` already has at 40 against 29.

The failure mode this removes is worth naming: the check goes red with the suite entirely green, which reads at a glance exactly like a real test failure — the same PR had a genuine one (#287) two attempts earlier.

### Found while reading the file

`test-with-code-coverage` has no `if: github.event_name != 'pull_request' || github.event.pull_request.draft == false` guard, so it runs on draft PRs while `test` skips them. That is a behaviour change, not a timeout fix, so it is not in this diff.

`fail-fast: false` on both jobs is #279 and untouched here.

### Verification

- [x] Reproduced first: run 34401975496 attempt 2, coverage job `cancelled` at 30m01s with all steps `success`; attempt 1 `failure` at 16m05s, attempt 3 `success` at 10m17s
- [x] No Julia code changed, so no test covers this — the verification is CI on this branch running the coverage job under the new cap
- [x] Local CI mirror (`agent ci-local`): PASS in 9m
- [x] GitHub CI on this PR: PASS, both runs. The `Julia 1.12` coverage job took **28m18s** — green, and 1m42s inside the cap it used to run under. The `1.11` coverage job took 21m54s; the plain `test` ubuntu jobs 27m11s and 25m43s.
- Docs n/a · no REUSE in this repo · branch cut from `main` at 3f75701, no conflict
- Changelog: n/a, a CI job cap is not user-visible
- Risk: 50 is generous enough that a genuinely hung suite now burns 50 minutes of runner time before it is cut, instead of 30.

### Scope

+1 / -1 in `.github/workflows/CI.yml`, one integer. No stack.

Closes #288 · task `VortexStepMethod.jl-288`

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -73,7 +73,7 @@ jobs:
           prefix: ${{ runner.os == 'Linux' && 'xvfb-run -a' || '' }}
   test-with-code-coverage:
     name: Julia ${{ matrix.julia_version }} - ${{ matrix.os }} - ${{ matrix.julia_arch }} - with code coverage
-    timeout-minutes: 30
+    timeout-minutes: 50
     strategy:
       fail-fast: false
       matrix:


1-Bort-1 (commented):

**Verdict: APPROVE WITH COMMENTS** · 0 inline, 0 off the diff

**Good**

- Raises only the coverage job timeout to 50, keeping the plain `test` job unchanged, matching the PR card's one-idea scope.
- The number is backed by data: cold coverage runs land at 26–28 minutes against a 30-minute cap, so 50 restores a safety margin comparable to the plain job's 40-minute cap.
- No Julia code, docs, or dependencies are touched; only `.github/workflows/CI.yml` line 76 changes.

**Not good**

- The coverage job still runs on draft PRs while the plain `test` job skips them; the PR card notes this but leaves the inconsistency in place.
- 50 minutes is generous enough that a genuinely hung suite now wastes more runner time before the runner kills it.

1-Bart-1 (approved):

Issue #291, opened 2026-09-09
The generated ram-air polar fixture is not reproducible, so every test asserting numbers on it draws a fresh geometry each run


1-Bort-1:
`ram_air_matrix_wing` (`test/test_data_utils.jl:26`) builds its wing by slicing `data/ram_air_kite/ram_air_kite_body.obj` and sweeping NeuralFoil over it into `test/generated/` (gitignored, so CI regenerates it every run). That fixture is not reproducible across machines, and not even across two attempts of the same job.

`norm(jac_fwd)` from the `@info "POLAR_MATRICES linearize jacobian norms"` line in `test/solver/test_forwarddiff.jl`, run 34401975496 on #285 — same commit, same fixture config (`n_panels=8`, `n_sections=4`, `alpha_range = deg2rad.(-5:5:15)`, `delta_range = deg2rad.(-3:3:3)`):

| where | `norm_fwd` |
|---|---|
| attempt 1, Julia 1.12 ubuntu with code coverage | 2.1143274312638014 |
| attempt 3, Julia 1.12 macOS aarch64 / 1.12 windows x64 / 1.11 ubuntu coverage | 2.114573062599269 |
| attempt 3, Julia 1.11 ubuntu / 1.12 ubuntu / 1.12 ubuntu coverage | 2.165929233230906 |
| agent box, Debian x64 | 2.2058861369294172 |

A 4.3% spread, and the *same job* — Julia 1.12 ubuntu with code coverage — gives 2.1143274312638014 on attempt 1 and 2.165929233230906 on attempt 3.

It is the fixture and not the code. The INVISCID sibling in the same file, which builds its wing in Julia and touches no generated file, reads `norm_fwd = 11.110746084832735`, `norm_fd = 11.110746104003324` on this box and **the same 17 digits** in that failing coverage job. So VSM's solver, geometry pipeline, ForwardDiff and FiniteDiff paths are bit-reproducible, coverage build included; what moves is what `obj_to_yaml` writes.

Two things are ruled out:

- **Not the `%.4f` polar rounding** (`write_polar_matrix_csv`, `src/airfoil_aero/airfoil_solvers/common.jl:62`). Jittering every Cl/Cd/Cm entry by ±1 unit in the last written decimal, 15 draws, moves `norm_fwd` by ±0.02% — two orders below the observed spread.
[...]

PR #292, opened 2026-09-09
test_forwarddiff.jl's POLAR_MATRICES check flakes at ~4%, and the oper

### LINKED ISSUE #287: test_forwarddiff.jl's POLAR_MATRICES check flakes at ~4%, and the oper

1-Bort-1:
### TL;DR

The POLAR_MATRICES check compares forward AD against a finite difference on a polar table that is piecewise-linear with knots every 5°, and its 1e-5 step spans a knot whenever a panel's converged alpha lands within ~3e-4° of one — which is how CI got 0.0400. Converging the LOOP solve to 1e-11 and taking the difference at 1e-8 keeps the secant inside one cell everywhere I could put it, exactly on a knot included, and the bound goes from 1e-3 to 1e-4 because the measured floor is now 7.1e-8.

### What is actually discontinuous

Nothing in VSM. The INVISCID sibling in the same file reads `norm_fwd = 11.110746084832735`, `norm_fd = 11.110746104003324` on this box and **the same 17 digits** in the ubuntu-1.12-coverage job that failed the POLAR check on #285 — so the solver, the geometry pipeline and both backends are bit-reproducible, coverage build and all.

What is discontinuous is the model the check differentiates. `build_interps` (`src/panel.jl:225`) hands `linear_interpolation` a 5-knot alpha grid, so `cl(alpha)` is a broken line, and a forward difference straddling one of those corners measures the average of two slopes while forward AD takes one of them. Bisecting the freestream alpha until panel 3's converged alpha sits on the 10° knot, everything else exactly as the test has it:

| panel 3 alpha − 10° | `rel_err` |
|---|---|
| −4.9e-3° | 3.35e-06 |
| −2.4e-4° | 4.52e-05 |
| −9.8e-5° | **6.88e-02** |
| −1.5e-5° | **1.23e-01** |
| +5e-15° | **8.83e-03** |
| +2.4e-5° | **2.29e-03** |
| +8.1e-5° | 5.97e-04 |
| +1.6e-4° | 3.22e-06 |

CI's 0.04001429189872729 sits inside that range, and 300x of margin sits just outside it. Whether a given run lands in the window is decided by the NeuralFoil fixture, which is redrawn every run and does not come out the same twice — that part is #291, and it is why this cannot be fixed by choosing a better operating point.

### The two knobs that decide how wide the window is

The step, obviously: the window scales with it. But also `rtol`, because the LOOP loop stops on `normalized_error < rtol` (`src/solver.jl:1080`), so the iteration count is a step function of the input and the converged alpha carries the leftover of whichever iteration it stopped on. Sampling CFz along theta_2 in 2000 steps of the test's own 1e-5, the one iteration-count change in that span moves the finite difference by 1.45e-4 — normalised, 1.05e-4, a tenth of the whole 1e-3 budget, from a step the test does not control. At `rtol=1e-11` that staircase is gone (worst jitter 5.3e-6, and that is curvature). Until the solve is that tight, the panel alpha itself wobbles by ~1e-5°, so shrinking the step alone does not help: it is the pair that works.

Scanned across the same knot, offset in degrees of freestream alpha:

| offset | as the test had it | rtol 1e-11, step 1e-7 | **rtol 1e-11, step 1e-8** |
|---|---|---|---|
| 1e-3° | 3.20e-06 | 3.46e-08 | 7.09e-08 |
| 3e-4° | **3.74e-02** | 3.44e-08 | 2.25e-08 |
| 1e-4° | **1.00e-01** | 3.40e-08 | 3.29e-08 |
| 1e-5° | **1.31e-01** | 3.42e-08 | 6.30e-08 |
| 3e-6° | **1.38e-01** | **2.07e-02** | 5.00e-08 |
| 1e-6° | **1.53e-01** | **8.35e-02** | 6.85e-08 |
| 1e-7° | **2.03e-01** | **1.12e-01** | 3.40e-08 |
| 0 | **8.83e-03** | **1.15e-01** | 5.47e-08 |

The chosen pair never leaves 2.2e-08..7.1e-08 anywhere on that scan. The bound follows: 1e-4 is 1400x above the measured floor and still ten times above the worst the exit staircase can contribute at this step, and it is a tightening, not a loosening.

### What I did not do

I did not move the operating point. #287 is right that 320e756's `defl_rad` reached nothing — `delta` is exactly 0.0 for every panel, but at an exact delta knot the alpha-direction slope is the same from either cell, and `delta` is not a differentiated input, so a delta knot cannot produce this. Moving the freestream alpha to maximise clearance from the alpha knots would buy about 1° of margin against a fixture that moves panel alphas by 0.1–0.2° per draw, which is a smaller factor than it sounds and would need moving again the next time the fixture moves.

The INVISCID testset keeps its 1e-5 step and 1e-3 bound: it has no table, so no knot, and it has never flaked.

`rtol=1e-11` costs 685 LOOP iterations against 390, out of `max_iterations=1500` — and 671–685 across ±10% on every Cl, twenty times the fixture's own spread, so the halve-the-relaxation retry stays far away; warm, the file's wall time is 0.1 s either way, so the CI figure for this testset is compilation. If #284 lands and `rtol` starts meaning the unrelaxed residual, every number here only improves.

### Verification

- [x] Reproduced first: unchanged file at the knot-adjacent operating point (12.810477774611°, panel 3 alpha 9.99995111°) → `Expression: rel_err < 0.001` / `Evaluated: 0.10032484514560956 < 0.001`
- [x] `test/solver/test_forwarddiff.jl` red before that change, green after at the same operating point (7/7), and green at the committed operating point (7/7, juliaserver, 49.7 s cold)
- [x] Local full suite: `agent ci-local` — see below · GitHub CI: see below
- [x] Up to date with `main` (branch parent is 3f75701) · docs not run — no public symbol touched · no REUSE lint in this repo · `jetls` not installed on this box
- Benchmark: n/a
- Risk: exactly on a knot the check still relies on forward AD and a forward difference choosing the same side of the corner; that held at every offset I measured, but it is a coincidence of convention rather than something the test asserts.

### Scope

+5 / −4 in `test/solver/test_forwarddiff.jl` and nothing else. Not stacked: #285 changes only `gamma_loop!`'s NONLIN branch. Opened #291 for the fixture that is redrawn every run, which this does not fix.

Closes #287 · task `VortexStepMethod.jl-287`

### DIFF
--- a/test/solver/test_forwarddiff.jl
+++ b/test/solver/test_forwarddiff.jl
@@ -59,13 +59,13 @@ using Test
         ram_solver = Solver(ram_body;
             aerodynamic_model_type=VSM,
             is_with_artificial_damping=false,
-            rtol=1e-7,
+            rtol=1e-11,
             solver_type=LOOP,
             use_gamma_prev=false,
         )
 
         v_a = 15.0
-        aoa_rad = deg2rad(7.5)  # off-grid (grid is every 1°) to avoid piecewise-linear node discontinuities
+        aoa_rad = deg2rad(7.5)
         y_op = [zeros(4);
                 [cos(aoa_rad), 0.0, sin(aoa_rad)] * v_a;
                 zeros(3)]
@@ -80,11 +80,12 @@ using Test
             ram_solver, ram_body, y_op;
             theta_idxs=1:4, va_idxs=5:7, omega_idxs=8:10,
             aero_coeffs=true,
-            backend=AutoFiniteDiff(absstep=1e-5, relstep=1e-5))
+            # A wider step lets the secant span a knot of the polar's 5° alpha grid.
+            backend=AutoFiniteDiff(absstep=1e-8, relstep=1e-8))
         @test conv_fd
 
         @info "POLAR_MATRICES linearize jacobian norms" norm_fwd=norm(jac_fwd) norm_fd=norm(jac_fd)
         rel_err = maximum(abs.(jac_fwd .- jac_fd)) / maximum(abs, jac_fwd)
-        @test rel_err < 1e-3
+        @test rel_err < 1e-4
     end
 end
--- a/test/solver/test_forwarddiff.jl
+++ b/test/solver/test_forwarddiff.jl
@@ -3,6 +3,8 @@ using DifferentiationInterface
 using LinearAlgebra
 using Test
 
+relative_error(jac, reference) = maximum(abs.(jac .- reference)) / maximum(abs, reference)
+
 @testset "ForwardDiff linearize" begin
     n_panels = 10
     span = 20.0
@@ -37,9 +39,8 @@ using Test
             backend=AutoFiniteDiff(absstep=1e-5, relstep=1e-5))
         @test fd_converged
 
-        @info "INVISCID linearize jacobian norms" norm_fwd=norm(jac_fwd) norm_fd=norm(jac_fd)
-        rel_err = maximum(abs.(jac_fwd .- jac_fd)) / maximum(abs, jac_fwd)
-        @test rel_err < 1e-3
+        @info "INVISCID jacobian norms" norm_fwd=norm(jac_fwd) norm_fd=norm(jac_fd)
+        @test relative_error(jac_fd, jac_fwd) < 1e-3
     end
 
     @testset "NONLIN+ForwardDiff is rejected" begin
@@ -79,13 +80,10 @@ using Test
         jac_fd, _, conv_fd = VortexStepMethod.linearize(
             ram_solver, ram_body, y_op;
             theta_idxs=1:4, va_idxs=5:7, omega_idxs=8:10,
-            aero_coeffs=true,
-            # A wider step lets the secant span a knot of the polar's 5° alpha grid.
-            backend=AutoFiniteDiff(absstep=1e-8, relstep=1e-8))
+            aero_coeffs=true, backend=nothing)
         @test conv_fd
 
-        @info "POLAR_MATRICES linearize jacobian norms" norm_fwd=norm(jac_fwd) norm_fd=norm(jac_fd)
-        rel_err = maximum(abs.(jac_fwd .- jac_fd)) / maximum(abs, jac_fwd)
-        @test rel_err < 1e-4
+        @info "POLAR_MATRICES jacobian norms" norm_fwd=norm(jac_fwd) norm_fd=norm(jac_fd)
+        @test relative_error(jac_fd, jac_fwd) < 1e-4
     end
 end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,12 @@
   The default is off: a post-stall solve that misses the tolerances still returns
   its `solver_status == FAILURE` solution.
 
+### Changed
+
+- `linearize`'s `fd_absstep`/`fd_relstep`, the finite-difference step it takes
+  when handed `backend=nothing`, default to `1e-8` instead of `1e-3` — the
+  `sqrt(eps)` scale the `NONLIN` Newton Jacobian already perturbs by.
+
 ## VortexStepMethod v5.0.0 2026-09-07
 
 ### Added
--- a/src/solver.jl
+++ b/src/solver.jl
@@ -1240,8 +1240,8 @@ deflections (one per unrefined section), apparent wind `(vx, vy, vz)`, and angul
 `(ωx, ωy, ωz)` respectively.
 
 `backend` accepts any `DifferentiationInterface` backend; `AutoForwardDiff()` (the default)
-requires `solver_type=LOOP`. `fd_absstep`/`fd_relstep` are forwarded only when the backend is
-`AutoFiniteDiff`.
+requires `solver_type=LOOP`. `backend=nothing` builds an `AutoFiniteDiff` from the
+`fd_absstep`/`fd_relstep` steps, which no other backend reads.
 
 Returns `(jac, results, converged)` where `results` is `(F, M, moment_unrefined_dist...)` —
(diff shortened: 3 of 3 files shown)


test/solver/test_forwarddiff.jl
```diff
@@ -80,11 +80,12 @@ using Test
             ram_solver, ram_body, y_op;
             theta_idxs=1:4, va_idxs=5:7, omega_idxs=8:10,
             aero_coeffs=true,
-            backend=AutoFiniteDiff(absstep=1e-5, relstep=1e-5))
+            # A wider step lets the secant span a knot of the polar's 5° alpha grid.
```
1-Bort-1, line 83: MINOR: The added comment describes the old, wider 1e-5 step, while the code below uses a narrower 1e-8 step; a future reader can invert the intent. Rephrase it to say the small step keeps the secant inside one polar cell, or delete it.

1-Bort-1 (commented):

**Good**

- Only test/solver/test_forwarddiff.jl is changed, matching the single-idea, minimal scope in the PR card.
- Tightens rtol to 1e-11 and shrinks the finite-difference step together, addressing the real discontinuity instead of loosening the bound.
- The test bound is tightened from 1e-3 to 1e-4, so the fix strengthens rather than weakens the assertion.
- The PR description includes the scan data and opens #291 for the redrawn fixture, so the remaining risk is tracked.

**Not good**

- `test/solver/test_forwarddiff.jl:83` — The added comment describes the old, wider 1e-5 step, while the code below uses a narrower 1e-8 step; a future reader can invert the intent. Rephrase it to say the small step keeps the secant inside one polar cell, or delete it.
- The comment at line 83 says a wider step spans a knot, but the new step is narrower; it will mislead the next reader.
- rtol=1e-11 and aoa=7.5 are driven by the polar-grid discontinuity, yet the code does not mention that relationship for readers who skip the PR description.

test/solver/test_forwarddiff.jl
```diff
@@ -80,11 +80,12 @@ using Test
...
+            backend=AutoFiniteDiff(absstep=1e-8, relstep=1e-8))
```
1-Bart-1: Now this is only changed in the step. Shouldn't this be the new default for the src code as well? @1-Bort-1

1-Bart-1 (approved):

test/solver/test_forwarddiff.jl
1-Bort-1: Done in 1a4a039: `linearize`'s `fd_absstep`/`fd_relstep` default to 1e-8 instead of 1e-3, and the test now passes `backend=nothing` rather than restating the step. Numbers behind the choice are in the PR body.

Issue #295, opened 2026-09-10
The Kulfan least-squares fit has no rank guard, so a crowded resample writes an airfoil at 1e4 scale with a normal residual


1-Bort-1:
`fit_kulfan_parameters(x, y, ::LeastSquaresFit)` solves `A \ y_norm` at `src/airfoil_aero/kulfan.jl:147` with no ridge, no rank or condition check, no bound on the coefficients, and no check on what `kulfan_to_coordinates` then evaluates from them. When the contour handed to it has its stations crowded into a narrow band of x, the upper Bernstein columns go numerically dependent and the solve returns weights five orders of magnitude too large — while reporting a perfectly ordinary residual, because those weights cancel on the stations they were fitted against.

On SK100 mesh section 16, measured on the same slice:

| | `cond(A)` | `max abs(coeff)` | `max abs(y)` of the resampled contour |
| --- | --- | --- | --- |
| stations spread normally | 1.073e3 | 1.23 | 0.0812 |
| stations crowded | 2.519e11 | 457400 | 13013 |

The residual is 0.0207 in both rows. `kulfan_to_coordinates` evaluates the weights on its own cosine grid rather than on the fitted stations, which is where the cancellation stops holding and the shape explodes.

`KulfanBasis` in `src/airfoil_aero/deform.jl:44` already ridge-regularises the same shape of solve — `(shape' * shape + ridge * scale * I) \ shape'` — and its docstring describes exactly this failure, so the package has the remedy in one place and not in the other.

### What crowds the stations

`resample_arc` (`src/airfoil_aero/shrink_wrap.jl:293`) spaces stations in a curvature-weighted measure whose density comes from `smoothed_curvature` (`:229`). Its only guard is `span > 0` at `:242`, and `pivot_contour` emits near-coincident vertices around a rounded convex corner, so a span of order 1e-16 gives a density of order 1e8. On section 16 the weighted measure reaches 103570 for a curve whose arclength is 1.063, with five nodes holding all of it; 118 of the 120 upper stations then land inside a band 0.03 wide. `enforce_min_spacing!` works in the same distorted measure and does not undo it.

### Why it matters beyond the section that triggered it

[...]

PR #301, opened 2026-09-11
write_section_aero writes an all-NaN deflected .dat, and read_section_

### LINKED ISSUE #300: write_section_aero writes an all-NaN deflected .dat, and read_section_
write_section_aero writes an all-NaN deflected .dat, and read_section_aero loads it as a NaN contour without complaint

When XFoil converges at no angle of attack for one deflection, `generate_airfoil_aero` leaves that whole contour column at its `fill(NaN, …)` initial value, `write_section_aero` writes it out as `airfoils/<id>_d<tag>.dat` full of `NaN NaN` rows, and `read_section_aero` then loads it back into the wing's contour interpolants without a word. The loaded `Wing` has a NaN airfoil shape at **every** deflection, including exactly 0°.

### The chain

`src/airfoil_aero/airfoil_solvers/xfoil_solver.jl:60-63` — a non-converged angle returns a `SectionSolution` with **empty** node arrays. (`NeuralFoilSolver` always fills `s.x = def.x` at `neuralfoil_solver.jl:32-42`, so this is XFoil-only.)

`src/airfoil_aero/section_aero_gen.jl:27-39` — the contour matrices start as `fill(NaN, n_node, n_delta)` and a column is written only from a solution whose `length(s.cp) == n_node`, the modal node count over the whole grid. Every angle of that deflection hitting the `continue` on line 32 leaves `x[:, jd]` entirely NaN. A column is all-finite or all-NaN, never mixed. `fill_node_nans!` repairs only `cp`/`cf`, never `x`/`y`.

`src/airfoil_aero/geometry_gen.jl:48-49` — `generate_airfoils` checks only column 1 (`all(isnan, clvals)` over `sols[1]`), so a deflection that converged nowhere does not stop the id being reported as solved.

`src/airfoil_aero/section_aero_gen.jl:137-141` — `write_section_aero` writes every non-zero deflection unconditionally, *before* the all-NaN guard on `:142-146` runs. That guard's own comment states the hazard the deflected files are not protected from:

```julia
    # never write `$dat_prefix.dat` all-NaN (read_dat_coordinates would drop it to empty)
    finite(jd) = any(isfinite, view(aero.x, :, jd))
```

### The two readers disagree, and the lenient one is the dangerous one

`read_dat_coordinates` (`src/airfoil_aero/airfoil_io.jl:12`) skips any line whose first character is not a digit, `-` or `.`, so it returns **zero points** and the caller crashes.

`read_dat` (`src/section_aero.jl:126-137`) has no such filter, and `tryparse(Float64, "NaN")` succeeds, so `read_section_aero` gets a full-length NaN vector, the lengths match, `x[:, jd] .= xd` succeeds, and the NaN reaches `x_interp`/`y_interp` (`section_aero.jl:82-85`). Linear interpolation between a finite and a NaN node is NaN everywhere, so `section_surface` returns a NaN contour at **every** delta — 0.0 included — and hands it to `live_polar.jl:136` and the Makie skin plot. No error, no warning.

### How it surfaced

CI job `Test end-user and developer setup`, twice on the same commit (runs [34585324837](https://github.com/OpenSourceAWE/VortexStepMethod.jl/actions/runs/34585324837)), on `examples/ram_air_kite.jl` with `delta_range = 0:1` and `XFoilSolver`. `Wing(yaml)` on line 74 loaded fine — through the lenient reader, with a NaN contour. Line 79's audit plot, which goes through the strict reader, is what actually failed:

```
LoadError: ArgumentError: reducing over an empty collection is not allowed
 [14] normalize_airfoil          @ src/airfoil_aero/kulfan.jl:71   (argmin on empty x)
 [15] fit_kulfan_parameters      @ src/airfoil_aero/kulfan.jl:136
 [17] generated_slices           @ ext/VortexStepMethodMakieExt.jl:1630
 [25] plot_slices_3d(path; delta=1.0, obj_path=...)
```

So the crash is the lucky case. The silent one is the default path for anyone loading a generated `geometry.yaml`.

Ten lines reproduce it with no solver at all — a `SectionAero` whose 1° column is NaN:

```
wrote: 7_d1.dat  isfile=true
NaN NaN
read back: 0 points
fit_kulfan_parameters threw: ArgumentError: reducing over an empty collection is not allowed
```

Reproducible on the GitHub runner, not on the agent box, where XFoil converges for all 8 solved sections of that mesh and every `_d1.dat` reads back 239 clean points. The setup-test workflow resolves dependencies fresh on each run (`test/test_setup.sh:62-65`) and the example, unlike `ram_air_matrix_dir` (`test/test_data_utils.jl:51-52`), does not pin BLAS threads — so which deflections converge moves between runs. Related: #291 on that fixture's instability.

### What I would do

1. Skip the column in `write_section_aero`'s deflection loop when `finite(jd)` is false, hoisting the helper above the loop — one condition, and it reuses the guard already sitting three lines below. No all-NaN `.dat` is ever written, so `isfile` remains the honest test for every reader.
2. Make `read_dat` and `read_dat_coordinates` agree on NaN rows. Two readers of one format with different tolerances is how the silent case got in; the strict one's behaviour is the one worth keeping.
3. Decide whether `generate_airfoils` should fail an id outright when a deflection converged at no angle, the way it already does for column 1 (`geometry_gen.jl:48-49`). That is a behaviour question rather than a bug, and it is the one I would want an opinion on.

No test asserts a deflected `.dat` is re-readable; the one round-trip testset (`test/airfoil_aero/test_airfoil_aero.jl:131`) builds an all-finite contour and round-trips through the lenient reader, so it could not have caught this.

Related to #295, which names the same `write_section_aero` guard — but that issue is about a degenerate *finite* fit at 1e4 scale, not an all-NaN column, and neither fix covers the other.

Found while working #298 (#299), whose red `Test end-user and developer setup` check this is. That diff only changes which filename `generated_slices` asks for, and for the `delta=1.0` this example uses the old and new schemes produce the identical `_d1.dat` — verified by evaluating both, and against the generated directory itself. Out of scope there.

1-Bort-1:
### TL;DR

`write_section_aero` no longer writes a deflection whose contour is all `NaN`, and the two `.dat` readers that disagreed about `NaN` rows are now one strict reader. A deflection the 2D solver converged at no angle used to reach a loaded `Wing` as a `NaN` airfoil shape at **every** deflection, 0° included, with no error and no warning.

### What was wrong

`generate_airfoil_aero` writes a contour column only from a solution whose node count matches the grid's modal one, and `XFoilSolver` returns empty node arrays for a non-converged angle — so a deflection that converged nowhere leaves `x[:, jd]` and `y[:, jd]` at their `fill(NaN, …)` initial value. `write_section_aero` wrote that column out as `airfoils/<id>_d<tag>.dat` full of `NaN NaN` rows, three lines above a guard whose comment already names the hazard for `{dat_prefix}.dat`.

Then the readers disagreed. `read_dat_coordinates` (submodule) kept only rows starting with a digit, `-` or `.`, so an all-`NaN` file read back as zero points and its caller crashed — that is the CI failure on #299, `argmin` on an empty `x` inside `normalize_airfoil`. `read_dat` (main package) had no such filter and `tryparse(Float64, "NaN")` succeeds, so `read_section_aero` got a full-length `NaN` vector, the lengths matched, and the `NaN`s landed in `x_interp`/`y_interp`. Linear interpolation between a finite node and a `NaN` node is `NaN` everywhere, which is why the contour came out `NaN` at 0° as well. Ten lines reproduce it with no solver at all, a `SectionAero` whose 1° column is `NaN`:

```
wrote _d1.dat: isfile=true
section | NaN NaN | NaN NaN
read_dat_coordinates: 0 points
read_section_aero: SectionAero{…}
  contour at delta=0.0: x=[NaN, NaN, NaN, NaN, NaN]
  contour at delta=1.0: x=[NaN, NaN, NaN, NaN, NaN]
```

### What I changed

`write_section_aero` hoists the `finite(jd)` helper it already had above the deflection loop and skips a column with no finite contour, so no all-`NaN` `.dat` is ever written and `isfile` stays the honest test for a generated deflection — which is what the Makie audit plot already assumes, warning and drawing nothing for a deflection it finds no file for.

The duplicate reader is gone rather than aligned. `read_dat_coordinates` moves into `src/section_aero.jl` beside `read_node_table` and `read_section_aero`, since loading lives in the main package and writing in `AirfoilAero`; `AirfoilAero` imports and re-exports it, so every existing caller and the `using VortexStepMethod.AirfoilAero: read_dat_coordinates` in the tests are unchanged. It now drops a row by value (`isfinite`) instead of by first character, which is what the first-character filter was approximating: that filter skipped `NaN NaN` only because `N` is not a digit, and let `0.5 NaN` through. I searched for further copies before deleting the second one and found a third — `read_dat_coords` in `test/airfoil_aero/test_airfoil_aero.jl`, a test-local reader of the same format — and deleted it too, in the file this PR already changes.

`read_section_aero` is the last link: a deflected contour that is missing, or that does not hold exactly its `Cp` table's node count, now warns and returns `nothing` instead of quietly filling the interpolants. `nothing` is a state the loader already returns and `add_section!` already takes, so an airfoil with a broken contour set loses its surface aero and keeps its polar, rather than poisoning the whole wing's geometry.

### The third item, and a better fix underneath it

Should `generate_airfoils` fail an id outright when a deflection converged at no angle, the way `all(isnan, clvals)` already fails one for column 1? I would, and for the same reason: the coefficients of such a column survive only because `fill_node_nans!` interpolated them from the neighbouring deflections, so what the id offers at that delta is a fabricated polar and, after this PR, no surface data at all. `reuse_valid_airfoils=true` then maps those sections to the nearest id that solved everywhere — a visible, documented degradation — and `reuse_valid_airfoils=false` errors, which is what a caller who turned the fallback off asked to hear. The push-back is that it costs a whole airfoil for one fussy deflection, and #291 says that fixture's convergence moves between runs.

There is a cheaper fix underneath it that I would rather see: the contour does not depend on the solver converging at all. It is `deform_section`'s own output, handed *in*, and `NeuralFoilSolver` already carries it straight through (`s.x = def.x`). I measured what XFoil gives back at `repanel=false` — 239 nodes against `def`'s 239, `max|dx| = max|dy| = 0.0` at both 0° and 3° — so `generate_airfoil_aero` could take `x`/`y` from `def` whenever its node count matches the modal one and no geometry column would ever be `NaN`, for either backend. That does not subsume this PR: a converged deflection whose node count is not the modal one still leaves a `NaN` column, so the writer's guard stays the backstop. Say the word and I will open it as its own task.

This is the fix for the red `Test end-user and developer setup` check that #299 carried, and that check is green here: with no all-`NaN` `_d1.dat` written, `generated_slices` takes its `isfile` branch and warns instead of crashing in `normalize_airfoil`. #295 names the same guard but is about a degenerate *finite* fit, and neither fix covers the other.

### The one red job is not this diff

`Julia 1.12 - windows-latest - x64` on `5e78ab6` is #287's flake, and `main` settles it rather than my say-so: twenty minutes after this run, `main`'s own push CI for the #308 merge ([run 34696079829](https://github.com/OpenSourceAWE/VortexStepMethod.jl/actions/runs/34696079829)) failed the identical assertion in `Julia 1.12 - ubuntu` at `relative_error(jac_fd, jac_fwd) = 0.0420681631196805`, against this branch's `0.04206403446037785` at `test/solver/test_forwarddiff.jl:87`. Same test, same ~0.042 branch of the bimodal jump #287 records, on a commit carrying none of these seven files. Which cell fires moves run to run and OS to OS: windows red and 1.12 ubuntu green here, 1.12 ubuntu red and windows green on `main`, and on the earlier `3c4fd7e` both 1.12 ubuntu cells at 0.03999947 and 0.04001102 with windows green. It is a bug in the test — a 1e-4 bound over a discrete interpolant jump — and #287 owns it.

So the failing check is not evidence about this diff, and I ruled the diff out by measurement rather than by argument. `test_forwarddiff` builds its wing from a generated directory, which does go through the reader this PR rewrites, so I deleted that fixture and regenerated it through this branch's writer: 16 `.dat` files, **not one with a `NaN` row**, and the strict reader returns coordinates identical to the lenient one it replaces on **all 16**. The change cannot alter what that test loads. On that fresh fixture the file is 7/7 green with `norm(jac_fwd) = 3.2971127281317214` and `rel_err = 5.22e-8` — 1900x inside the bound, and 800000x below what CI saw.

### Merging `main` in

`main` has since taken #294, #299, #306, #310 and #312, and v5.1.0 went out. Only `CHANGELOG.md` conflicted, both sides having added to the same `### Fixed` block: my entry moves to `## Unreleased`, where it belongs now that v5.1.0 shipped without it, and both of `main`'s entries stay verbatim.

One conflict was semantic and the textual merge hid it. #294 added `@testset "generate_airfoils fits the wrapped contour it is handed"` at the end of `test/airfoil_aero/test_airfoil_aero.jl`, and it calls `read_dat_coords` — the test-local third reader this PR deletes. The merged file was clean to `git` and `UndefVarError` to Julia. Both calls now go through `read_dat_coordinates`, which is the point of the deletion; the two files it reads are finite, so the strict reader returns the same coordinates.

`main`'s #312 guards the other end of the same bug — the audit plot skipping a deflected `.dat` that reads back empty — and the two do not overlap: #312 protects a directory already generated with an all-`NaN` `.dat`, this stops one being written. Its "Audit slices (Makie)" testset writes its blank file by hand, so it still exercises that guard with this writer in place, and passes 7/7 here.

### Verification

- [x] Reproduced first: `contour at delta=0.0: x=[NaN, NaN, NaN, NaN, NaN]` from a `SectionAero` whose only bad column is 1°, written and read back through the public writer and loader
- [x] `test/airfoil_aero/test_airfoil_aero.jl` — new testset "a deflection with no contour is neither written nor loaded as NaN" 2 passed / 3 failed before the change (the `.dat` was written, the load returned a `SectionAero`), 7/7 after
- [x] Re-run against the merged tree (juliaserver, examples env): `airfoil_aero/test_airfoil_aero.jl` (87 over 11 testsets, #294's new one among them at 2/2 and this PR's at 7/7), `airfoil_aero/test_live_polar.jl` (93), `obj_adapter/test_obj_adapter.jl` (52, the generate → `Wing(yaml)` round trip), `solver/test_backend_comparison.jl` (15), `yaml_geometry/test_yaml_geometry.jl` (178), `plotting/test_plotting.jl` (58 + 19 + #312's 6 and 7, the other caller of the moved reader), `Aqua.jl` (10) — no failures
- [x] Docs build clean against the merged tree (`docs/make.jl`, only the pre-existing `size_threshold_warn` notices on `private_functions.md` and `functions.md`) · REUSE: n/a, no `bin/reuse_lint` · merged up to `origin/main` (`8b0c6bc`)
- [x] GitHub CI on `5e78ab6`: six of seven checks green — Documentation, `Julia 1.11 - ubuntu`, `Julia 1.12 - ubuntu`, `Julia 1.12 - macOS-aarch64`, codecov/patch and `Test end-user and developer setup`. `Julia 1.12 - windows` red on #287's flake alone (`0.04206403446037785`), with this PR's testset 7/7 inside it, and `main` red on that same assertion at `0.0420681631196805` twenty minutes later. Local CI mirror (`agent ci-local`, one matrix cell) on the merged tree: PASS in 13m
- [x] `test/solver/test_forwarddiff.jl` on this branch, fixture deleted and regenerated through the new writer: 7/7 green, `norm(jac_fwd) = 3.2971127281317214`, `rel_err = 5.22e-8` against the 1e-4 bound (juliaserver, examples env)
- [x] Every `test/generated/` fixture deleted and rebuilt through this branch's writer, then the two suites that consume them re-run, because the copies my earlier runs used were written before this branch took #294: `obj_adapter/test_obj_adapter.jl` 52/52 and `plotting/test_plotting.jl` 58 + 19 + 6 + 7 — same counts as against the cached fixtures, so nothing in this PR's evidence rested on the stale contour
- Risk: `read_section_aero` returning `nothing` where it used to throw a `SystemError` for a deflected contour it cannot open. Every generated directory on disk today with an all-`NaN` `_d1.dat` now warns and loads without surface aero instead of loading a `NaN` one, which is the intent — but a hand-made directory whose deflection tag does not match `delta_suffix` degrades where it used to fail loudly.

### Scope

+70 / −65 across 7 files against the merge base (`8b0c6bc`), of which 20 deleted lines are the two duplicate `.dat` readers and 23 are the moved one. No stack.

Found on the way, not fixed here: `test/generated/`'s fixture path is keyed on `hash((n_sections, alpha_range, delta_range))` and nothing about the code that fills it (`test/test_data_utils.jl:46-49`), so a box whose cache predates a contour-generation change keeps testing the old contour. It cost me a wrong number: my first `test_forwarddiff` run on this merged tree reported `norm(jac_fwd) = 2.2059`, #294's pre-fix doubly-wrapped value, from a directory generated the day before; regenerating gave `3.29711`. CI never sees it, checking out fresh each run. Keying the path on the generator's version, or stamping the config into the directory, is a change to shared test infrastructure that no test here needs — its own task, and worth one to whoever picks up #287, whose next step is reproducing that fixture.

Also found, not fixed: `test/test_body_aerodynamics.jl`, `test/test_solver.jl` and `test/test_yaml_geometry.jl` are tracked and zero bytes — leftovers of the move into `test/<area>/`, which nothing includes. `include`ing one silently does nothing, which is how I met them. Three file deletions in a diff that has no other reason to open them is a `cleanup:` PR; say the word and I will open it.

Closes #300 · task `VortexStepMethod.jl-300`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,12 @@
   `test/solver/solver_test_wing.yaml` wing at 26.6° it stopped 3.6% below `LOOP`'s
   peak circulation and reported `FAILURE`, and now lands on the same distribution
   with a fixed-point residual at machine precision.
+- A deflection the 2D solver returned no contour for is no longer written as an
+  all-`NaN` airfoil `.dat`: `write_section_aero` skips that column, so `isfile` stays
+  the honest test for a generated deflection. `read_section_aero` warns and returns
+  `nothing` for a contour that does not hold its `Cp` table's nodes, where it used to
+  read the `NaN`s back and interpolate a `NaN` airfoil shape at *every* deflection,
+  0° included.
 
 ## VortexStepMethod v5.0.0 2026-09-07
 
--- a/docs/src/private_functions.md
+++ b/docs/src/private_functions.md
@@ -116,7 +116,7 @@ window_alpha
 assemble_polar_matrix
 load_matrix_polar_data
 read_aero_matrix
-read_dat
+read_dat_coordinates
 read_node_table
 write_node_rows
 convert_node_table
@@ -203,7 +203,6 @@ sigmoid
 create_2d_polars
 lei_poly_coeffs
 resolve_airfoil
-read_dat_coordinates
 write_dat
 write_polar_csv
 write_polar_matrix_csv
--- a/src/airfoil_aero/AirfoilAero.jl
+++ b/src/airfoil_aero/AirfoilAero.jl
@@ -9,8 +9,8 @@ using Xfoil
 using Printf: @sprintf
 using ..VortexStepMethod: SectionAero, interpolate_matrix_nans!, delta_suffix,
                           write_node_rows, section_surface, set_polar!,
-                          KulfanParameters, calculate_cl, calculate_cd,
-                          calculate_cm
+                          read_dat_coordinates, KulfanParameters, calculate_cl,
+                          calculate_cd, calculate_cm
 
 include("kulfan.jl")
 include("deform.jl")
--- a/src/airfoil_aero/airfoil_io.jl
+++ b/src/airfoil_aero/airfoil_io.jl
@@ -1,27 +1,4 @@
 """
-    read_dat_coordinates(path) -> (x, y)
-
-Read airfoil coordinates from a Selig-format `.dat` file, skipping header and
-comment lines.
-"""
-function read_dat_coordinates(path::String)
-    x = Float64[]
-    y = Float64[]
-    for line in eachline(path)
-        s = strip(line)
-        (isempty(s) || !(isdigit(s[1]) || s[1] == '-' || s[1] == '.')) && continue
-        parts = split(s)
-        length(parts) >= 2 || continue
-        xp = tryparse(Float64, parts[1])
-        yp = tryparse(Float64, parts[2])
-        (xp === nothing || yp === nothing) && continue
-        push!(x, xp)
-        push!(y, yp)
-    end
-    return x, y
-end
-
-"""
     write_dat(filepath, name, x, y) -> filepath
 
 Write airfoil coordinates to a Selig-format `.dat` file.
--- a/src/airfoil_aero/section_aero_gen.jl
+++ b/src/airfoil_aero/section_aero_gen.jl
@@ -119,7 +119,9 @@ end
 
 Write a [`SectionAero`](@ref) as human-readable files. The airfoil contours share
 `dat_prefix`: `{dat_prefix}.dat` (contour at `delta=0`) plus
-`{dat_prefix}_{delta_suffix(δ)}.dat` per non-zero deflection. The per-node `Cp`/`cf`
+`{dat_prefix}_{delta_suffix(δ)}.dat` per non-zero deflection. A deflection whose
+contour is all `NaN` is skipped, and `{dat_prefix}.dat` falls back to the first
+deflection that has one. The per-node `Cp`/`cf`
 tables share `table_prefix` (defaults to `dat_prefix`): `{table_prefix}_cp.{ext}` and
 `{table_prefix}_cf.{ext}`, where `ext` is `table_format`, either `:csv` (default,
 readable) or `:arrow` (binary, an order of magnitude faster to load). Pass a separate
@@ -134,15 +136,15 @@ function write_section_aero(dat_prefix::AbstractString, aero::SectionAero;
         throw(ArgumentError("table_format must be :csv or :arrow, got :$table_format"))
     mkpath(dirname(dat_prefix))
     mkpath(dirname(table_prefix))
+    finite(jd) = any(isfinite, view(aero.x, :, jd))
     for (jd, d) in enumerate(aero.delta_range)
-        iszero(d) && continue
+        (iszero(d) || !finite(jd)) && continue
         write_dat("$(dat_prefix)_$(delta_suffix(d)).dat", "section",
                   aero.x[:, jd], aero.y[:, jd])
     end
-    # never write `$dat_prefix.dat` all-NaN (read_dat_coordinates would drop it to empty)
-    finite(jd) = any(isfinite, view(aero.x, :, jd))
     jz = findfirst(iszero, aero.delta_range)
-    jdat = jz !== nothing && finite(jz) ? jz : findfirst(finite, eachindex(aero.delta_range))
+    jdat = jz !== nothing && finite(jz) ? jz :
+           findfirst(finite, eachindex(aero.delta_range))
     write_dat("$dat_prefix.dat", "section", aero.x[:, jdat], aero.y[:, jdat])
     cp_path = "$(table_prefix)_cp.$table_format"
     cf_path = "$(table_prefix)_cf.$table_format"
--- a/src/section_aero.jl
+++ b/src/section_aero.jl
@@ -118,21 +118,24 @@ function delta_suffix(delta)
 end
 
 """
-    read_dat(path) -> (x, y)
+    read_dat_coordinates(path) -> (x, y)
 
 Read Selig `.dat` airfoil coordinates (two whitespace-separated columns), skipping the
-name/header line and any non-numeric lines.
+name/header line, comments, and any row that is not a finite pair. The single `.dat`
+reader; [`write_dat`](@ref VortexStepMethod.AirfoilAero.write_dat) is its writer.
 """
-function read_dat(path::AbstractString)
+function read_dat_coordinates(path::AbstractString)
     x = Float64[]
     y = Float64[]
-    for ln in eachline(String(path))
-        p = split(strip(ln))
-        length(p) >= 2 || continue
-        xv, yv = tryparse(Float64, p[1]), tryparse(Float64, p[2])
-        (xv === nothing || yv === nothing) && continue
-        push!(x, xv)
-        push!(y, yv)
+    for line in eachline(String(path))
+        fields = split(strip(line))
+        length(fields) >= 2 || continue
+        xp = tryparse(Float64, fields[1])
+        yp = tryparse(Float64, fields[2])
+        (xp === nothing || yp === nothing) && continue
+        (isfinite(xp) && isfinite(yp)) || continue
+        push!(x, xp)
+        push!(y, yp)
     end
     return x, y
 end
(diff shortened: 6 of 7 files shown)


1-Bort-1:
**CI: `Julia 1.12 - windows-latest - x64 - pull_request` failed** on `5e78ab662`.

https://github.com/OpenSourceAWE/VortexStepMethod.jl/actions/runs/34696347903/job/103560385346

1-Bart-1 (approved):

Issue #313, opened 2026-09-12
cleanup: the generate_airfoils changelog entry sits under the released


1-Bort-1:
cleanup: the generate_airfoils changelog entry sits under the released v5.1.0 header

#294 merged from `4b261f0` while v5.1.0 was being cut, so `CHANGELOG.md` now files

    - `generate_airfoils` fits the contour it is handed rather than shrink-wrapping it a second time. …

under `## VortexStepMethod v5.1.0 2026-09-11`, a release that shipped without it. It belongs in the `## Unreleased` section that already exists above it. `bin/release` reads the top section, so leaving it there puts the fix in the notes of a tag that does not contain it.

Two commits on `agent/BeyondTheSim.jl-17` already do exactly this and did not make the merge — OpenSourceAWE/VortexStepMethod.jl@5a4b7ec moves the entry, OpenSourceAWE/VortexStepMethod.jl@a820826 folds the two-line inline comment in `test/airfoil_aero/test_airfoil_aero.jl` into one line (CLEAN_CODE.md §3). Cherry-pick both off that branch.

+6 / -8, no code and no test behaviour.

1-Bort-1:
Nothing to cherry-pick: #314 already is those two commits. It is open on the same branch `agent/BeyondTheSim.jl-17`, targets `main`, is not a draft, and its whole diff against `main` is exactly this move — OpenSourceAWE/VortexStepMethod.jl@5a4b7ec putting the entry back under `## Unreleased` and OpenSourceAWE/VortexStepMethod.jl@a820826 folding the comment in `test/airfoil_aero/test_airfoil_aero.jl`. `git diff origin/main...refs/pull/314/head --stat` gives +7 / -8 across those two files and nothing else, byte-identical to what a cherry-pick here would produce.

So merging #314 is what closes this. It reads MERGEABLE, review required, CI re-running. A branch of my own would be the same diff twice and would conflict with #314 on whichever merged second, so I pushed nothing and opened no PR.

If you would rather have the changelog straight before #314 is reviewed, say so and I will open the standalone `cleanup:` PR off `main` with just those two commits — then #314 drops to no diff at all.

PR #314, opened 2026-09-12
Fit the contour generate_airfoils is handed, not a second wrap of it

1-Bort-1:
The red check was not this branch's, and it has since gone green on its own. `test/solver/test_forwarddiff.jl:87`, testset "AutoForwardDiff matches AutoFiniteDiff (LOOP, POLAR_MATRICES)", evaluated `relative_error(jac_fd, jac_fwd) = 0.04207 < 1e-4` on `4b261f0`. Re-running that same job on that same commit, without one line changed, turned it green — and with it all seven checks. The same testset had already failed twice on `main` before this branch existed, at 0.04001 (OpenSourceAWE/VortexStepMethod.jl@6af7183) and 0.040003 (OpenSourceAWE/VortexStepMethod.jl@fb1eea7). It is #287's flake, which #292 closed without removing; I reopened it with everything measured below.

Chasing it turned up one thing this change owes the reader. Fitting the contour `generate_airfoils` is handed drops the wingtip panel incidence at that test's operating point from **19.04/18.93 deg to 14.99/14.94 deg**, and `norm(jac_fwd)` from 2.21 to 3.30. The doubly-wrapped fit had been pushing the tips 4 deg past the polar table's top knot into `extrap_flat`, where `dcl/dalpha` is zero and a tip carries no alpha sensitivity at all. That is the fix working, on a wing this diff never touches.

It is not what makes the check flake, and neither is the finite-difference step #292 shrank. `rel_err` stays in 3e-8…5e-8 across solver `rtol` 1e-11…1e-7 crossed with steps 1e-8…1e-4 — flat, so nothing is trading truncation against solver noise. Walking the operating point across the knot the tips now sit under changes nothing either: 268 points (alpha 0:0.25:20 at zero sideslip, then alpha 5:0.5:10 crossed with beta -4:0.5:4) give a worst `rel_err` of **2.3e-7**. What is left is the fixture, #291 — the failing runner reported `norm(jac_fwd) = 3.28551` where this box gets `3.29711` on the same commit, so it was not generating the same NeuralFoil tables. That axis cannot be walked from here.

One loose end this merge left behind: v5.1.0 was cut while the branch was open, so `CHANGELOG.md` on `main` now files this fix under `## VortexStepMethod v5.1.0 2026-09-11`, a release that shipped without it. The two commits that put it back under `## Unreleased` (OpenSourceAWE/VortexStepMethod.jl@5a4b7ec, and @a820826 folding a two-line inline comment into one) landed on the branch after the merge. Queued as #313 to cherry-pick.

### Verification

- [x] CI: re-run of `Julia 1.12 - ubuntu-latest - x64 - pull_request` on `4b261f0`, no code change — green, all seven checks green on that commit
- [x] `test/airfoil_aero/test_airfoil_aero.jl` green with `main` merged in, incl. `generate_airfoils fits the wrapped contour it is handed | 2 2` (juliaserver, tables regenerated)
- [x] `test/solver/test_forwarddiff.jl` green here, 7/7, `rel_err = 4.84e-8` — 2000x inside the bound CI missed
- Risk: the ~0.040 jump is still unexplained, so the check can come back red on a runner whose generated tables differ. Nothing here makes that likelier — `main`'s doubly-wrapped fit measures 4.80e-8 on this box against this branch's 4.84e-8.

Task `VortexStepMethod.jl-294`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,16 @@
 # Changelog
 
+## Unreleased
+
+### Fixed
+
+- `generate_airfoils` fits the contour it is handed rather than shrink-wrapping it a
+  second time. Both adapters wrap before they call, with the `wrap_method` their
+  settings name, so the second wrap discarded that setting and degenerated the Kulfan
+  fit: on the SK100's 45 mesh sections nine came out with `.dat` coordinates at 1e2 to
+  1e4 instead of 0..1, and any consumer mapping the mesh onto a structure rejected the
+  geometry.
+
 ## VortexStepMethod v5.1.0 2026-09-11
 
 ### Added
@@ -19,12 +30,6 @@
 
 ### Fixed
 
-- `generate_airfoils` fits the contour it is handed rather than shrink-wrapping it a
-  second time. Both adapters wrap before they call, with the `wrap_method` their
-  settings name, so the second wrap discarded that setting and degenerated the Kulfan
-  fit: on the SK100's 45 mesh sections nine came out with `.dat` coordinates at 1e2 to
-  1e4 instead of 0..1, and any consumer mapping the mesh onto a structure rejected the
-  geometry.
 - The `NONLIN` solver backtracks along each Newton step instead of always taking
   it whole, so it converges past stall where the full step used to cycle: on the
   `test/solver/solver_test_wing.yaml` wing at 26.6° it stopped 3.6% below `LOOP`'s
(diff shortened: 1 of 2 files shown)


1-Bort-1 (commented):

**Good**

- Single focused change: removes the second shrink-wrap from generate_airfoils and nothing else.
- Tests the fix it ships: new testset directly checks that written coordinates match the once-fitted contour within 1e-4.
- Changelog entry is behavior-facing, compact, and names the defect (coordinates at 1e2..1e4 instead of 0..1).
- No new public symbols, no docstrings needing reachability, no manifest or version bumps.

**Not good**

- Test at line 287-288 could compare extrema tuples directly instead of collect() .- collect(), but it is still readable.

1-Bart-1 (approved):

PR #320, opened 2026-09-14
Refuse a contour XFoil's panel code cannot take

### LINKED ISSUE #322: analyze_sweep on an XFoilSolver can end the Julia process, exit code 0 and no exception
`analyze_sweep(::XFoilSolver, definition, alphas, reynolds_number)` can end the Julia process instead of returning. There is no exception, no stacktrace and the exit code is **0**, so in a REPL it takes the session away and in a script it looks like a successful run that printed half its output. No `try` in the caller can catch it: it is a bare Fortran `STOP` inside XFoil, and gfortran exits 0 on one that carries no message.

### How it shows up

Splitting the call on a contour that does it, `Xfoil.set_coordinates(def.x, def.y)` returns normally (`n=239 nb=239`) and `Xfoil.solve_alpha(0.0, reynolds_number; reinit=true)` never returns. Nothing on stdout, nothing on stderr, `echo $?` says 0.

The contours that do it are the ones whose upper and lower surfaces cross, which is what `deform_section`'s re-wrap makes of a thin already-wrapped section (#321). Nine shapes from one mesh section of a kite wing, one per rolling-ball radius, all 239 nodes, each run in its own process:

| `min_concave_radius` | shortest panel | neighbour ratio | max turn | crossing panel pairs | XFoil |
| --- | --- | --- | --- | --- | --- |
| 0.05 | 8.03e-4 | 2.20 | 45.0° | 34 | **ended the process** |
| 0.10 | 7.38e-4 | 4.14 | 48.3° | 34 | returned |
| 0.15 | 2.08e-4 | 16.65 | 180.0° | 60 | **ended the process** |
| 0.20 | 2.66e-4 | 14.45 | 180.0° | 58 | **ended the process** |
| 0.30 | 3.09e-4 | 13.54 | 180.0° | 48 | **ended the process** |
| 0.40 | 4.07e-4 | 3.54 | 180.0° | 29 | returned |
| 0.60 | 3.82e-4 | 6.29 | 180.0° | 29 | returned |
| 0.80 | 3.90e-4 | 5.53 | 180.0° | 35 | returned |
| the stored `.dat`, one wrap of the same cloud | 5.01e-4 | 7.50 | 32.3° | **0** | returned |

The rows at 0.05 and 0.10 are the point: they agree on every column and disagree on life and death. So which of these XFoil walks into a `STOP` on is not something the caller — or a guard — can predict from the geometry. What is true of all eight is that none of them is a simple closed curve, which is the precondition a panel method actually has; the one contour with zero crossings is the only valid input in the set.

### What I would do

Check the contour in `analyze_sweep` before handing it to XFoil and throw, so a section XFoil has no answer for is one failed sweep the caller can catch rather than the end of the session. #320 does that on branch `agent/BeyondTheSim.jl-53`, with the geometry above as its test.

Not the same concern as #274, which is XFoil's `__init__` cost, nor #321, which is why the crossing contour is produced in the first place.

Reported as 1-Bart-1/BeyondTheSim.jl#53, where an example script lost its process this way after a fresh install.

1-Bort-1:
### TL;DR

`analyze_sweep(::XFoilSolver, ...)` checks the contour before handing it to XFoil and throws `ArgumentError`, because a self-crossing contour can reach a bare Fortran `STOP` in there and end the Julia process with exit code 0 and no exception. The same check catches a contour with more nodes than XFoil's panel arrays hold, which XFoil refused while leaving the previously loaded airfoil in place, so every angle came back solved on the wrong shape.

### What was wrong

Found from 1-Bart-1/BeyondTheSim.jl#53: an example script ended its Julia process partway through, silently, exit code 0, no stacktrace — in a REPL it takes the session away, in a script it looks like a successful run that printed half its output. Bisecting the call, `Xfoil.set_coordinates` returns normally and `Xfoil.solve_alpha(0.0, Re; reinit=true)` never does. Exit 0 with nothing on stdout or stderr is a gfortran `STOP` with no message.

The contours that do it come out of `deform_section`, whose re-wrap of an already-wrapped thin section produces a curve whose upper and lower surfaces cross. That is not an airfoil, and a panel method has no solution for it, so which of them terminates the process and which returns is not something the caller can predict — of nine shapes sliced from one mesh section at nine rolling-ball radii, four returned and five took the process with them, while every geometric statistic I measured (node count, shortest panel, neighbouring-panel ratio, maximum turn, crossing count) put survivors and killers in the same range:

| wrap radius | shortest panel | neighbour ratio | max turn | crossings | XFoil |
| --- | --- | --- | --- | --- | --- |
| 0.05 | 8.03e-4 | 2.20 | 45.0° | 34 | **killed the process** |
| 0.10 | 7.38e-4 | 4.14 | 48.3° | 34 | returned |
| 0.15 | 2.08e-4 | 16.65 | 180.0° | 60 | **killed the process** |
| 0.20 | 2.66e-4 | 14.45 | 180.0° | 58 | **killed the process** |
| 0.30 | 3.09e-4 | 13.54 | 180.0° | 48 | **killed the process** |
| 0.40 | 4.07e-4 | 3.54 | 180.0° | 29 | returned |
| 0.60 | 3.82e-4 | 6.29 | 180.0° | 29 | returned |
| 0.80 | 3.90e-4 | 5.53 | 180.0° | 35 | returned |
| the stored `.dat`, one wrap of the same cloud | 5.01e-4 | 7.50 | 32.3° | **0** | returned |

The rows at 0.05 and 0.10 are the point: they agree on every column and disagree on life and death. So the guard cannot be a predictor of the `STOP`. What separates the one safe contour from the rest is that it is the only simple closed curve in the set, and that is the precondition a panel method actually has — which is what `validate_xfoil_contour` now enforces.

I do not claim this makes `analyze_sweep` proof against every contour XFoil might `STOP` on. It ends this class, which is the one a shrink wrap can produce, and it converts an unpredictable process death into a failure the caller can catch.

### The second defect in the same function

While bisecting I found that a contour with more than `Xfoil.IQX - 5` = 281 nodes is refused by XFoil's `ABCOPY`, which prints to stdout and **returns**, leaving the previously loaded airfoil in the panel arrays. `analyze_sweep` never looked, so it solved the sweep on whatever shape was loaded before and returned it as this one. Feeding a NACA 0012 at 299, 339, 399, 499 and 569 nodes gave `cl = 0.5604994052186286` every time — the 279-node answer, to the last digit. Same function, same idea (say when XFoil did not take the contour), so it rides here.

### Where I would push back

The real bug is upstream of this: `deform_section`'s re-wrap should not produce a self-crossing contour in the first place, and until it stops doing so these sections have no XFoil polar at all rather than a poor one. That is the shrink wrap's own concern and would change what every generated polar means, so it is #321 rather than this diff.

The contour test is O(n²) in the node count — 239 nodes is ~28k segment pairs, run once per sweep against a viscous march of many angles. I did not benchmark it; it is not on any hot path I can find, and `compare_live_polar`'s per-panel call is still one XFoil sweep per check.

### Verification

- [x] Reproduced first: `Xfoil.set_coordinates` ok, then `solve_alpha(0.0, Re; reinit=true)` never returns — process gone, exit 0, no Julia error and no Fortran message
- [x] `test/airfoil_aero/test_airfoil_aero.jl` red before, green after (5/5 new assertions, 108/108 file, exit 0). Red was run with the source change stashed: the folded contour returned instead of throwing, and the 399-node contour printed `Maximum number of panel nodes : 281 / Current airfoil cannot be set.` and still returned a solution
- [x] The nine real contours above now raise `ArgumentError` naming the crossing panels, all nine **in one process** where each killer previously needed its own; the valid one still runs
- [x] `test/solver/test_backend_comparison.jl` PASS 15/15 (42.5 s) — the suite's only XFoil march · `test/airfoil_aero/test_live_polar.jl` PASS 79/79
- [x] Docs build clean (exit 0, no missing docstring) · up to date with `origin/main` at 23f129b
- [ ] Full local suite: not run · GitHub CI: pending
- REUSE lint: n/a, this repo has no `bin/reuse_lint`
- Benchmark: n/a
- Risk: the guard rejects contours that today happen to return — four of the nine above. Their answers were a panel solution on a curve that crosses itself, so I judge that no loss, but it is a behaviour change for any caller that was using one.

### Scope

+96 / -1 across 5 files: the check and its two geometry helpers in `airfoil_aero/`, one regression testset, one changelog entry under `## Unreleased`, four `@docs` entries. The regression test uses a synthetic folded Kulfan section rather than the mesh section that found this — that geometry is in a private repo and this one is public. Closes #322.

From 1-Bart-1/BeyondTheSim.jl#53 · task `BeyondTheSim.jl-53`


1-Bart-1:
Merge main into this branch

1-Bart-1:
Fix the failing test and conflicts @1-Bort-1

1-Bort-1:
Picked up as `VortexStepMethod.jl-320`, on the branch this pr already has.

1-Bort-1:

1-Bort-1:
The description above is this branch as it stands.

### Commits

- `2637859` Test and support Julia 1.12 and 1.13, drop 1.11
- `0576422` Changelog: Julia 1.12 and 1.13
- `2b312ea` Merge pull request #327 from OpenSourceAWE/agent/326-julia-1-13-vortexstepmethod-jl
- `f63f7b6` Test solve! on a body built from two wings
- `dcaf811` Zero the damping correction when smooth_circulation! skips smoothing
- `d3cf02f` Pin lei_poly_coeffs to the Python Breukels regression output
- `deac49a` Test against the Python verification cases' references
- `4078a92` Test that moments scale as N·m when the geometry scales
- `78bba14` Compare the solve! solutions directly and changelog the unit labels
- `e971c74` Give untyped empty containers their element type
- `9ef9f10` Build a Solver from VSMSettings or from panel and section counts
- `47d8f19` Rename the *_array locals to _dist, after the per-panel fields
- `be131ba` Give the renamed AIC and results docstrings their full signatures
- `7f2d15f` Build every flat inviscid test wing with one helper
- `1fac603` Merge pull request #334 from OpenSourceAWE/agent/93-add-test-cases-from-the-python-verificat
- `226fe70` Merge pull request #335 from OpenSourceAWE/agent/209-add-tests-that-check-if-moment-have-righ
- `a95bfac` Merge pull request #338 from OpenSourceAWE/agent/147-refactoring-part-iii
- `b83d4c1` Merge pull request #339 from OpenSourceAWE/agent/143-add-test-for-multiple-wings
- `a932ee1` Merge remote-tracking branch 'origin/main' into agent/153-remove-the-parameter-body-aero-from-the-
- `246237a` Merge pull request #341 from OpenSourceAWE/agent/336-rename-the-array-locals-to-say-whether-t
- `d165c7f` Deprecate the Solver constructors that take a body_aero
- `fc3daab` Name the apparent wind va, va_vec and va_dist in src locals and arguments
- `245ab96` Truncate the SVD in the least-squares Kulfan fit
- `30b0756` Share one straight-segment kernel between bound and trailing vortices
- `90afa39` Name the apparent wind va, va_vec and va_dist in tests, examples and docs
- `4965a5f` Add the Gaunaa et al. 2024 spanwise-flow viscous drag correction as a solver option
- `701adef` Wrap the renamed lines at 92 columns
- `524103b` Finish the va rename in error messages, test helpers and alignment
- `9af8d69` Turn the body about a reference point in set_va!, and apply yaw_rate
- `e185eec` Plot cl, cd and cm against alpha per panel for every aero model
- `de35935` Share one panel coefficient path between the polar plots
- `580cd2f` linearize and make_dual_shadow take a body with several wings
- `1d76212` Name linearize's locals and share the two-wing test setup
- `1c493cf` Merge pull request #352 from OpenSourceAWE/agent/332-add-the-optional-spanwise-flow-viscous-d
- `0aefbe8` Merge branch 'main' into agent/295-the-kulfan-least-squares-fit-has-no-rank
- `bf7246f` Merge origin/main into agent/329-set-va-rotates-the-body-about-the-origin
- `4ccdea6` Merge origin/main into agent/344-linearize-and-make-dual-shadow-accept-a-
- `597339d` Merge origin/main into agent/153-remove-the-parameter-body-aero-from-the-
- `ff2900d` Merge origin/main: fold the va names into the viscous drag correction
- `0cc53b2` Merge remote-tracking branch 'origin/main' into agent/331-plot-section-polars-covers-only-polar-ve
- `9ae2ab2` List plot_section_polars under Added in the changelog
- `574b15b` Give each wing's unrefined-section range one source and test the angle split
- `465d199` Test both vortex functions for azimuthal velocity in one testset
- `3bef8ce` Merge pull request #328 from OpenSourceAWE/agent/53-add-unit-tests-for-uncovered-functions
- `83867a3` Merge pull request #349 from OpenSourceAWE/agent/337-one-name-per-apparent-wind-quantity-va-v
- `28d4e11` Merge origin/main into agent/153-remove-the-parameter-body-aero-from-the-
- `6ecd77a` Merge origin/main: carry the va_vec names into the reference_point change
- `0b28330` Merge origin/main into agent/344-linearize-and-make-dual-shadow-accept-a-
- `c02a2ab` Merge origin/main: carry the va rename into the shared vortex kernel
- `3dc74fc` Fold the second Unreleased "Fixed" changelog heading into the first
- `f8761aa` Fix tests
- `8a2b61d` Merge branch 'main' into agent/329-set-va-rotates-the-body-about-the-origin
- `e24e1e4` Merge pull request #355 from OpenSourceAWE/agent/344-linearize-and-make-dual-shadow-accept-a-
- `086ee32` Merge pull request #351 from OpenSourceAWE/agent/333-velocity-3d-trailing-vortex-projects-ont
- `7103702` Merge remote-tracking branch 'origin/main' into agent/331-plot-section-polars-covers-only-polar-ve
- `a035f39` Merge origin/main: make_dual_shadow builds every wing and keeps the reference point
- `1a80247` Merge remote-tracking branch 'origin/main' into agent/295-the-kulfan-least-squares-fit-has-no-rank
- `12c11f4` Merge origin/main: build the two-wing linearize test solver from counts
- `da666bf` Build the billowing example's solvers with Solver(settings)
- `b725371` Warn when the least-squares Kulfan fit drops singular values
- `2665240` Merge pull request #354 from OpenSourceAWE/agent/331-plot-section-polars-covers-only-polar-ve
- `7724c0d` Merge pull request #353 from OpenSourceAWE/agent/329-set-va-rotates-the-body-about-the-origin
- `4cd7a7e` Merge origin/main: keep the count constructors under set_va!'s reference point
- `0c94c6d` Merge pull request #340 from OpenSourceAWE/agent/153-remove-the-parameter-body-aero-from-the-
- `abd1176` Merge remote-tracking branch 'origin/main' into agent/BeyondTheSim.jl-53
- `9d0f44e` Merge pull request #342 from OpenSourceAWE/agent/295-the-kulfan-least-squares-fit-has-no-rank
- `fefa7a6` Merge remote-tracking branch 'origin/main' into agent/BeyondTheSim.jl-53
- `d1ffa0c` Test the panel that closes the contour back to node 1

[Review just these changes](https://github.com/OpenSourceAWE/VortexStepMethod.jl/pull/320/files/b7a47bf899f2b4a8f40bf939454ac5ae22847f90..d1ffa0c03f094a49905755b380a3f4b92fe63ada)

1-Bort-1:
Merged in fefa7a6, with the follow-up merge of #342 in the same push; `CHANGELOG.md` was the only conflict. The PR reads MERGEABLE and all 7 checks are green.

1-Bort-1:
Merged `origin/main` into the branch in fefa7a6; the only conflict was `CHANGELOG.md`, where this branch's Fixed bullet moved under main's `## Unreleased`. No test change was needed: the red check was `AutoForwardDiff matches AutoFiniteDiff (LOOP, POLAR_MATRICES)` on Julia 1.12 / windows at 0.0421, which is #360 — fixed on main on 2026-09-17, and the branch was 60 commits behind it. All 7 checks are green now, windows included, and the local full suite passes 6718/6719 (the 1 is the pre-existing broken).

src/airfoil_aero/airfoil_solvers/common.jl
```diff
@@ -116,6 +116,41 @@ function deform_section(x, y, delta; crease_frac=0.9, thickness_frac=1.0,
...
+    crossing_panels(x, y) -> Tuple{Int,Int} or nothing
+
+The first pair of non-neighbouring panels of the closed contour `(x, y)` that
+cross, or `nothing` when the contour is a simple closed curve.
+"""
+function crossing_panels(x, y)
+    nodes = collect(zip(x, y))
+    last_panel = length(nodes) - 1
+    for i in 1:last_panel, j in (i + 2):last_panel
+        i == 1 && j == last_panel && continue
```
1-Bort-1: MINOR: `crossing_panels` assumes a wrap-around closed contour: it never tests the closing segment node n → node 1 and it skips the pair (1, last_panel). That holds for `deform_section` output (`shrink_wrap` ends at `px[1]`, so first == last), but `compare_live_polar` builds its section straight from `kulfan_to_coordinates`, which leaves first ≠ last for any non-zero `TE_thickness` — there panels 1 and n-1 are genuinely non-neighbouring and the trailing-edge segment is untested, so a fold at the TE passes the guard and can still reach the `STOP`. For a truly closed contour the skip is dead anyway, since the shared node makes the strict test false.

1-Bort-1 (commented):

**Good**

- Single chokepoint: grep shows no `Xfoil.*` call site outside `xfoil_solver.jl`, and `analyze_section` delegates to `analyze_sweep`, so the guard cannot be bypassed.
- `Xfoil.IQX = 286` confirmed in the installed package (`Xfoil/n7wnc/src/Xfoil.jl:12`), so the 281-node bound matches XFoil's `ABCOPY` refusal the card describes.
- Order is right: validating before `set_coordinates` is the only place it works, since a refused contour leaves the previous airfoil loaded and `pane` runs after `set_coordinates`.
- No duplication: grepped `intersect|cross|orient|signed|det` over `src` — the repo had no segment-intersection or side-of-line helper, only plane crossings in `obj_slice.jl`.
- Regression test throws before any Fortran call, so it is fast and cannot itself kill the process; it covers both classes plus a clean control contour.
- Docstrings state only what, no why; CHANGELOG entry sits under `Unreleased/Fixed` at the neighbours' length; all four new private names are added to `private_functions.md` so docs stay complete.
- Scope matches the card (5 files, guard + 3 helpers + 1 testset + changelog + docs), tree clean, `.agent/plan.md` untracked.

**Not good**

- `src/airfoil_aero/airfoil_solvers/common.jl:148` — `crossing_panels` assumes a wrap-around closed contour: it never tests the closing segment node n → node 1 and it skips the pair (1, last_panel). That holds for `deform_section` output (`shrink_wrap` ends at `px[1]`, so first == last), but `compare_live_polar` builds its section straight from `kulfan_to_coordinates`, which leaves first ≠ last for any non-zero `TE_thickness` — there panels 1 and n-1 are genuinely non-neighbouring and the trailing-edge segment is untested, so a fold at the TE passes the guard and can still reach the `STOP`. For a truly closed contour the skip is dead anyway, since the shared node makes the strict test false.
- `src/airfoil_aero/airfoil_solvers/xfoil_solver.jl:62` — Throwing here aborts batches that were built to degrade instead: `generate_airfoil_aero` loops deltas and takes the modal node count specifically so "a single degenerate wrap cannot drop every other column", and `compare_live_polar`'s docstring promises a bad reference "comes back `NaN` rather than throwing, so a sweep over panels does not stop at the first one that fails". One self-crossing delta or panel now fails the whole grid or panel sweep; the card notes the behaviour change but not that these two callers document the opposite contract.
- The three geometry helpers sit in `airfoil_solvers/common.jl` (shared-backend file) although their only caller is the XFoil backend.
- Testset name says "XFoil refuses" while XFoil is never reached — the refusal is entirely ours.
- `side_of_line(a, b, p) -> Float64` in the docstring, but the function is generic in its argument types.
- No end-to-end assertion here that a valid contour still sweeps; that is left to `test/solver/test_backend_comparison.jl`.
- The card's disclosed behaviour change (four of nine contours that used to return now throw) gets no caller-side handling anywhere.

src/airfoil_aero/airfoil_solvers/xfoil_solver.jl
```diff
@@ -37,8 +54,12 @@ reinit at each side for convergence. Each converged angle reads the surface pres
 (`Xfoil.cpdump`) and the boundary layer (`Xfoil.bldump`, giving `cf` and the node
 coordinates) at the same panel nodes. Non-converged angles yield empty node arrays and
 `NaN` confidence.
+
+Throws `ArgumentError` for a contour XFoil's panel code cannot take, see
+[`validate_xfoil_contour`](@ref).
 """
 function analyze_sweep(solver::XFoilSolver, def::DeformedSection, alpha_range, Re)
+    validate_xfoil_contour(def)
```
1-Bort-1, line 62: MINOR: Throwing here aborts batches that were built to degrade instead: `generate_airfoil_aero` loops deltas and takes the modal node count specifically so "a single degenerate wrap cannot drop every other column", and `compare_live_polar`'s docstring promises a bad reference "comes back `NaN` rather than throwing, so a sweep over panels does not stop at the first one that fails". One self-crossing delta or panel now fails the whole grid or panel sweep; the card notes the behaviour change but not that these two callers document the opposite contract.

src/airfoil_aero/airfoil_solvers/common.jl
1-Bort-1: Fixed in d1ffa0c. `crossing_panels` now appends node 1 when the contour does not already end on it, so the trailing-edge panel is tested, and the `(1, last_panel)` skip is gone — once the two ends share a node it could never fire anyway. You are right about the path: `kulfan_to_coordinates` at `TE_thickness=0.02` gives first `(1.0, 0.01)` and last `(1.0, -0.01)`, and that is what `compare_live_polar` hands the solver. `crossing_panels([0,1,1,2],[0,2,-2,1])` — a contour whose only crossing is the closing panel — returned `nothing` before and returns `(2, 4)` after; that pair is the new assertion, with a TE-gap Kulfan section beside it to pin that a real open contour is still accepted.

src/airfoil_aero/airfoil_solvers/xfoil_solver.jl
1-Bort-1: Not changed, on both counts. `generate_airfoil_aero` already has the caller-side handling: `generate_airfoils` wraps it in try/catch at `src/airfoil_aero/geometry_gen.jl:41-75` and, with the default `reuse_valid_airfoils=true`, warns and reuses the nearest valid airfoil — so one self-crossing delta costs that airfoil, not every column. The modal node count next to it is tolerating a column that converged at an odd node count, which is a different thing from what this class did before the guard: it ended the process, so there was no degrading to preserve. `compare_live_polar`'s docstring promises NaN for a reference that will not converge anywhere on the ramp, and that is untouched — a contour that is not a simple closed curve is not a convergence failure. Swallowing it per delta or per panel is a decision about what a silently dropped column means to whoever reads the table, and I would not take it inside this diff.

1-Bart-1:
> Exit 0 with nothing on stdout or stderr is a gfortran STOP with no message.

This sounds like we need an upstream PR on Xfoil.jl as well for this.

1-Bort-1:
**Question from `VortexStepMethod.jl-320`**

Yes, but the fix belongs one repo below Xfoil.jl. The silent exit is the bare `STOP` in XFoil's `IBLPAN` (`byuflowlab/xfoil_light`, `src/xpanel.f:1432`; its two WRITE lines are commented out, so it prints nothing). It fires when the longer boundary-layer side plus the wake exceeds `IVX = 229` stations. On a self-crossing contour the stagnation point lands next to the trailing edge, so one side takes almost all N nodes. It cannot fire at 202 nodes or fewer. Predicting it from the inviscid solution was right in 9 of 9 synthetic runs: a figure-eight and a rear-swapped NACA 0012 die at 220-250 nodes and return NaN at 160-200. Xfoil.jl only ccalls a prebuilt `xfoil_light_jll`. Both repos are byuflowlab (tier C), and xfoil_light has had no commits since 2022.

Which route? A) The real fix: in xfoil_light, replace that `STOP` with an error flag in a COMMON block (and the `src_cs` copy too), rebuild `xfoil_light_jll` on Yggdrasil, then an Xfoil.jl PR that reads the flag and throws. That is three PRs and months of waiting. B) An Xfoil.jl-only PR: `solve_alpha` throws when `N + min(N÷8+2, 36) > IVX`. No rebuild, and it provably covers this STOP, but it refuses every viscous solve over 202 nodes, healthy contours included, so upstream may well say no. C) Nothing upstream: a follow-up VSM PR that caps `XFoilSolver`'s raw contours at 202 nodes, or repanels them to `npan = 160`. That fixes our users now. #320 stays as it is under all three.

My recommendation: A, with an issue on xfoil_light first, and C now in VSM so our users are not waiting on the release chain.

1-Bart-1:
Solving the cause rather than the effect

1-Bort-1:
**Question from `VortexStepMethod.jl-320`**

byuflowlab/xfoil_light is not a repo this box may push to.
Task `VortexStepMethod.jl-320` (OpenSourceAWE/VortexStepMethod.jl#320 — Refuse a contour XFoil's panel code cannot take) wants to fork it, open a pull request and file an issue there.
Thumbs up, or reply yes, and it goes ahead. Anything else and it does not.

Asked on Slack, and answered there.

1-Bort-1:
Opened in the repository this belongs to:

- Pull request: byuflowlab/xfoil_light#1

1-Bort-1:
Fixed at the cause upstream in 1-Bort-1/xfoil_light@63a3065, with the pull request against byuflowlab/xfoil_light opened from it. `IBLPAN` now checks the bound before writing and returns the point unconverged instead of the silent `STOP`. The three contours that ended the process now return `converged=false` (real and complex-step library); the other six are bit-identical. Xfoil.jl needs no change; this needs a Yggdrasil rebuild of `xfoil_light_jll` after merge. #320 is unchanged.

1-Bort-1:
**Local full suite: FAIL** (0 min, Julia 1.13.0, one cell of the matrix)

```
unknown function (ip: (nil)) at (unknown file)
unknown function (ip: 0x7fb3ae75e9ee) at /lib/x86_64-linux-gnu/libc.so.6
unknown function (ip: 0x7fb3ae753667) at /lib/x86_64-linux-gnu/libc.so.6
unknown function (ip: 0x7fb3ae7536ac) at /lib/x86_64-linux-gnu/libc.so.6
epoll_pwait at /lib/x86_64-linux-gnu/libc.so.6 (unknown line)
uv__io_poll at /workspace/srcdir/libuv/src/unix/linux.c:1404
uv_run at /workspace/srcdir/libuv/src/unix/core.c:430
ijl_task_get_next at /cache/build/builder-amdci4-4/julialang/julia-ci/src/scheduler.c:533
wait at ./task.jl:1296
wait_forever at ./task.jl:1196
jfptr_wait_forever_65680.1 at /home/agent/.julia/juliaup/julia-1.13.0+0.x64.linux.gnu/lib/julia/sys.so (unknown line)
jl_apply at /cache/build/builder-amdci4-4/julialang/julia-ci/src/julia.h:2318 [inlined]
start_task at /cache/build/builder-amdci4-4/julialang/julia-ci/src/task.c:1285
unknown function (ip: (nil)) at (unknown file)
Allocations: 7078566 (Pool: 7076302; Big: 2264); GC: 8
2026-09-21T12:25:22Z  KILLED by SIGTERM after 4m
ERROR: The Project.toml of the package being tested must have a name and a UUID entry
Stacktrace:
  [1] pkgerror(msg::String)
    @ Pkg.Types ~/.julia/juliaup/julia-1.13.0+0.x64.linux.gnu/share/julia/stdlib/v1.13/Pkg/src/Types.jl:68
  [2] test(ctx::Pkg.Types.Context, pkgs::Vector{PackageSpec}; coverage::Bool, test_fn::Nothing, julia_args::Cmd, test_args::Cmd, force_latest_compatible_version::Bool, allow_earlier_backwards_compatible_versions::Bool, allow_reresolve::Bool, kwargs::@Kwargs{io::IOContext{IO}})
    @ Pkg.API ~/.julia/juliaup/julia-1.13.0+0.x64.linux.gnu/share/julia/stdlib/v1.13/Pkg/src/API.jl:578
  [3] kwcall(::@NamedTuple{io::IOContext{IO}}, ::typeof(Pkg.API.test), ctx::Pkg.Types.Context, pkgs::Vector{PackageSpec})
    @ Pkg.API ~/.julia/juliaup/julia-1.13.0+0.x64.linux.gnu/share/julia/stdlib/v1.13/Pkg/src/API.jl:562
  [4] test(pkgs::Vector{PackageSpec}; io::IOContext{IO}, kwargs::@Kwargs{})
    @ Pkg.API ~/.julia/juliaup/julia-1.13.0+0.x64.linux.gnu/share/julia/stdlib/v1.13/Pkg/src/API.jl:172
  [5] test(pkgs::Vector{PackageSpec})
    @ Pkg.API ~/.julia/juliaup/julia-1.13.0+0.x64.linux.gnu/share/julia/stdlib/v1.13/Pkg/src/API.jl:161
  [6] test(; name::Nothing, uuid::Nothing, version::Nothing, url::Nothing, rev::Nothing, path::Nothing, mode::PackageMode, subdir::Nothing, kwargs::@Kwargs{})
    @ Pkg.API ~/.julia/juliaup/julia-1.13.0+0.x64.linux.gnu/share/julia/stdlib/v1.13/Pkg/src/API.jl:189
  [7] test()
    @ Pkg.API ~/.julia/juliaup/julia-1.13.0+0.x64.linux.gnu/share/julia/stdlib/v1.13/Pkg/src/API.jl:178
  [8] top-level scope
    @ none:1
  [9] eval(m::Module, e::Any)
    @ Core ./boot.jl:489
 [10] exec_options(opts::Base.JLOptions)
    @ Base ./client.jl:310
 [11] _start()
    @ Base ./client.jl:577
```

PR #325, opened 2026-09-16
Deprecate ObjAdapter's copy of the mesh inertia integral; SymbolicAWEModels owns it

### LINKED ISSUE #324: Bring RamAirKite.jl onto SymbolicAWEModels v0.18: Delete the duplicated inertia integral from VortexStepMethod
`center_to_com!`, `calculate_inertia_tensor` and `calc_inertia_y_rotation` live in
`src/obj_adapter/obj_geometry.jl` (`:246`, `:302`, `:341`) and are exported from
`ObjAdapter.jl:48`. SymbolicAWEModels carries its own copies
(`center_of_mass` and `calculate_inertia_tensor` in `src/obj_adapter.jl`,
`calc_inertia_y_rotation` in `src/rigid_body.jl:191`) and uses only
`ObjAdapter.read_faces` from here. Two implementations of one inertia integral in
two packages means nothing will ever report that they diverged.

Delete the three functions, drop them from the export list, delete their tests in
`test/ram_geometry/test_kite_geometry.jl` and `test/obj_adapter/test_obj_adapter.jl`,
and remove them from `docs/src/private_functions.md:247-248`. `read_faces` stays.
Nothing inside VortexStepMethod calls them, and a code search of OpenSourceAWE and
1-Bart-1 (RamAirKites, V3Kite, BeyondTheSim, SymbolicAWEModels) found no importer
on 2026-09-16; repeat that search before deleting. Removing public exports is
breaking: `BREAKING:` entry in the changelog.

Unit `vsm-delete` of plan `plans/ramairkite-obj-yaml-migration.md`, tracked in 1-Bart-1/Agents#401.
https://github.com/1-Bart-1/Agents/blob/eed96c809195812611065fddac64130fd4d419c9/plans/ramairkite-obj-yaml-migration.md

1-Bort-1:
### TL;DR

`center_to_com!`, `calculate_inertia_tensor` and `calc_inertia_y_rotation` stay in `ObjAdapter` but now emit a deprecation warning pointing at SymbolicAWEModels, and are to be removed in the next breaking release. SymbolicAWEModels computes mesh mass properties with its own copies, so two copies of one inertia integral live in two packages; deprecating rather than deleting keeps this release non-breaking, as asked in review.

### What changes for a caller

Nothing, apart from the warning. With `--depwarn=yes` (what `Pkg.test` and CI use) each call site warns once: `` `center_to_com!` is deprecated and will be removed in the next breaking release; SymbolicAWEModels computes mesh mass properties.`` With Julia's default `--depwarn=no` it is silent. The three docstrings say they are deprecated, `calc_inertia_y_rotation` gets the docstring it never had and joins the other two on `private_functions.md`. The changelog entry under `## Unreleased` → `### Changed` is a deprecation, not `BREAKING:`.

The message is built in one helper, `depwarn_inertia(name)`, so the three functions cannot word it differently.

### Who uses them

Nobody outside this repo. `gh search code` on OpenSourceAWE and 1-Bart-1, plus a shallow clone and grep of the repos where code search came back empty, found no import or call of VortexStepMethod's copies. SymbolicAWEModels (`src/obj_adapter.jl:50`, `src/system_structure/rigid_body.jl:191`) and StableAWEModels (`rigid_body.jl:180`) define their own, and SymbolicAWEModels takes only `ObjAdapter.read_faces` from here. BeyondTheSim `examples/neuralfoil.jl:67` does `using VortexStepMethod.ObjAdapter` and never calls them. So the warning should reach nobody, and removal in the next breaking release should break nobody either.

### Tests

The existing tests of the three functions stay and now wrap one call of each in `@test_deprecated`, so they assert the warning as well as the result. In `test_kite_geometry.jl` the "Interpolation Creation" testset also serialized an `_info.bin` of inertia tensor, COM and interpolants that nothing read; that block is gone, and with it `Serialization` from `test/Project.toml`. The `.obj` it writes stays, because "Alignment to principal frame" reads it back.

### Found on the way, not changed here

- `find_circle_center_and_radius` and `create_interpolations`, the other two `ObjWing` leftovers in `obj_geometry.jl`, have no caller in `src/` or `ext/`: only their own testsets call them. They are unexported, and deleting them wants its own importer search; with the aim of keeping breaking changes down, they could be deprecated alongside these three in a follow-up.

### Verification

- [x] Reproduced first: n/a, a deprecation. Red before: `test_kite_geometry.jl` under `--depwarn=yes` against main's `obj_geometry.jl` fails 3 × `Log Test Failed … Log Pattern: (:warn, r"deprecated"i, Ignored(), :depwarn)` (21 pass, 3 fail, 1 broken).
- [x] Green after, same run with this branch: 24 pass, 1 broken (the existing `@test_skip`), exit 0.
- [x] `test/obj_adapter/test_obj_adapter.jl`: 52/52 pass (juliaserver, test env, default depwarn)
- [x] Local full suite on e344c8e: PASS (7 min, Julia 1.12.7, one matrix cell, `agent ci-local`)
- [x] GitHub CI on e344c8e: 7/7 pass, Documentation included
- Docs build: not run locally, the `docs/` environment has no manifest; the CI Documentation job builds it
- No merge conflict with current main @ 246237a, which now carries #338's typing of `read_faces` (`git merge-tree`)
- REUSE lint: n/a, the repo does not use REUSE
- Benchmark: n/a
- Risk: a test suite run with `--depwarn=error` somewhere downstream would now error on these calls; the search above found no caller for that to hit.

### Scope

+29 / −17 across 5 files against main: the helper, three `depwarn_inertia` calls, docstring lines, one `private_functions.md` line, the changelog entry, three `@test_deprecated` wrappers, and the dead `_info.bin` block with its `Serialization` dependency. Most open PRs also add to `CHANGELOG.md`, and #320, #340, #346, #351 and #352 to `docs/src/private_functions.md`; #350 edits `test_obj_adapter.jl`. Whichever merges later may need a small textual merge.

Closes #324 · task `VortexStepMethod.jl-324`


1-Bort-1 (commented):

**Good**

- The diff matches the card: all three functions are removed along with their export, tests and docs entry, and `read_faces` stays. The commit stat is +11/−213 across 7 files, as the card says.
- No uses remain: grepping `src/`, `test/`, `docs/` and `*.toml` finds no reference to `center_to_com!`, `calculate_inertia_tensor` or `calc_inertia_y_rotation`. Only the untracked `.agent/` notes mention them.
- `using NonlinearSolve` is still needed because `find_circle_center_and_radius` calls `NewtonRaphson` at `obj_geometry.jl:89`. `LinearAlgebra` is still needed for `norm` at `obj_geometry.jl:211` and in `obj_slice.jl`.
- It is safe to drop `Serialization` from `test/Project.toml`: no test uses `serialize` now, and the root `Project.toml:31` still lists it, so the manifest does not change.
- The cleanup in the "Interpolation Creation" testset keeps behaviour: the `faces`, `.obj` write and `_info.bin` it removes are not used later in the testset. The `read_aero_matrix` and `create_interpolations` assertions are unchanged.
- No test was weakened: the only removed assertions tested the deleted functions, and the existing `@test_skip` is untouched.
- This is removal the rubric asks for: §2 wants one source per equation and nothing kept in case, and the other copy lives in SymbolicAWEModels, which takes only `read_faces` from here.
- The `BREAKING:` changelog entry goes under a new `## Unreleased`, and the version in `Project.toml` is not bumped.

**Not good**

- The changelog line says the names are no longer exported and then "they are gone", which says the same thing twice. "`ObjAdapter` drops `center_to_com!`, …" would be enough.
- The docs build was not run locally, so the CI Documentation job is the only check. The risk is low: only two `@docs` entries were removed, and `calc_inertia_y_rotation` had no docstring or docs entry.
- `test_obj_path` is now written and read only by the "OBJ File Reading" testset. It could move inside that testset, but that is optional.

1-Bort-1:
**Local full suite: PASS** (5 min, Julia 1.13.0, one cell of the matrix)

CHANGELOG.md
```diff
@@ -4,6 +4,9 @@
 
 ### Changed
 
+- BREAKING: `ObjAdapter` no longer exports `center_to_com!`, `calculate_inertia_tensor`
```
1-Bart-1: Keep the functions, mark them deprecated, and rm them in the next breaking release, because otherwise we have too many breaking changes.

1-Bort-1: Done in e344c8e: the three functions are back with their export, tests and docs, each emits a deprecation warning, and the changelog entry is a deprecation instead of BREAKING.

1-Bart-1: Ok fine we are going breaking already anyways. Just mark them breaking and not backwards compatible.

1-Bart-1: So remove them anyways.

1-Bort-1: Done in 8880556: the changelog entry is BREAKING again.

1-Bort-1: Removed in 8880556, reverting the deprecation round.

1-Bart-1 (approved):

PR #327, opened 2026-09-16
Test and support Julia 1.12 and 1.13, drop 1.11

### LINKED ISSUE #326: Julia 1.13: VortexStepMethod.jl
The shared procedure. Two matrices name minors — `CI.yml`'s main one plus its
`include:` entries, and a second job — and `setup-test.yml` is on `'1'`. Compat
says `julia = "1.10, 1.11"` though CI already tests 1.12; it becomes
`"1.12, 1.13"` like everywhere else. Both `Manifest-v1.11` and `Manifest-v1.12`
`.default`s are tracked.

Unit `vsm` of plan `plans/julia-1-13.md`, tracked in 1-Bart-1/Agents#282.
https://github.com/1-Bart-1/Agents/blob/bd9d5bf4f48b8dd00a025774b80762f9f2114c08/plans/julia-1-13.md

1-Bort-1:
### TL;DR

CI now tests Julia 1.12 and 1.13, `[compat]` says `julia = "1.12, 1.13"`, and a `Manifest-v1.13.toml.default` resolved under 1.13.0 takes the place of the 1.11 one. This is unit `vsm` of 1-Bart-1/Agents `plans/julia-1-13.md`; before this, compat said `"1.10, 1.11"` while CI already tested 1.12.

### What this repo had, of the five steps

**Matrix.** Only the `test` job in `CI.yml` names minors: its `version` list goes from 1.11/1.12 to 1.12/1.13. The three `include:` entries (ubuntu coverage, windows, macOS) name 1.12, which stays, so they don't change. The thread mentions a second job naming minors, but there isn't one: the `docs` job uses `setup-julia`'s default and `setup-test.yml` is on `'1'`, so both stay as they are. The job carries `fail-fast: false`; the plan puts that in its own `cleanup:` PR, not this one.

**Compat.** `julia = "1.10, 1.11"` → `"1.12, 1.13"`. There is no `SHA` entry, so the 1.13 `SHA` 1.0 bump needed nothing.

**Manifests.** Minimal resolve, not widened. Re-resolving the 1.12 `.default` under 1.12.7 moved one line: the package's own recorded version, stale at 4.0.0, is now 5.1.1. The 1.13 `.default` was seeded from that and resolved under 1.13.0. Only stdlibs and their JLLs moved (`SHA` 0.7→1.0, `LinearAlgebra`/`Pkg`/`SparseArrays` 1.12→1.13, `LibCURL`, `OpenBLAS_jll`, `SuiteSparse_jll`, …, plus `PCRE2_jll` added). No registry package changed. The 1.11 `.default` is deleted. It is half the size of the new one because 1.11 has no `[workspace]`, so it never held the `test`/`docs`/`examples` deps.

**Scripts and prose.** `bin/install`'s selector, its supported-version check, the Julia it installs when none is found and its manifest cleanup list all move to 1.12/1.13. `bin/update_default_manifests` moves the same way. The branch that copied the main manifest into each sub-project only ran on 1.11 (no workspaces there), so it is deleted instead of renamed. `README.md` said "Julia 1.10 or later" and `docs/src/index.md` said "1.11 or later"; both now say 1.12.

### Found on the way

- `bin/install` runs `juliaup default` when no `julia` is on the PATH. §2 of the agent rules says the scripts never do that. I only changed the version it names; fixing the behaviour is a separate `cleanup:`.
- Step 3 of the plan says `JULIAUP_CHANNEL=1.13 juliaserver launch .`, but that still came up on 1.12.7 here. `juliaserver` starts the session with `tmux new-session`, which takes its environment from the tmux server that is already running, not from the command that launched it. I restarted Julia inside this worktree's own pane as `julia +1.13` instead. The plan's other units will hit the same thing. (`juliaserver` also prints `md5: command not found` / `bc: command not found` on every call on this box.)
- `test/airfoil_aero/test_live_polar.jl`'s `refresh_allocs` docstring explains a workaround with "on Julia 1.11 `@allocated` boxes a `Float64`". 1.11 is no longer supported, but the test still passes on 1.13 as written, so I left it alone.
- `CHANGELOG.md` gets the same `## Unreleased` header #325 adds, so whichever merges second has a trivial conflict.

### Verification

- [x] Reproduced first: n/a — version move, nothing to reproduce
- [x] `./bin/install -y +1.13`: exit 0. All projects instantiated on the first try, precompile OK, Makie extension loaded, `settings/test_settings.jl` smoke test passed
- [x] `./bin/install -y +1.12`: exit 0, same steps
- [x] Julia 1.13.0 session, `test` env: `Aqua.jl` + `solver/test_solver.jl` + `airfoil_aero/test_live_polar.jl`, 137/137 pass
- [x] Local full suite (`agent ci-local`, Julia 1.12.7 cell): PASS in 10 min · GitHub CI: starts when the PR opens
- [x] Docs build (`docs/make.jl`, Julia 1.12.7): clean, apart from the size warnings on `private_functions.md` and `functions.md` that `main` already has
- [x] Up to date with `main` (23f129b) · no REUSE in this repo
- Risk: the macOS/Windows jobs stay on 1.12, so 1.13 is tested on ubuntu only.

### Scope

+3149 / −1478 across 10 files. The manifests account for 3122 added and 1451 deleted. The rest is 2 lines in `CI.yml`, 2 in `Project.toml`, 36 changed in `bin/install` (6 of them the deleted 1.11 branch), and one line each in `bin/update_default_manifests`, `README.md` and `docs/src/index.md`, plus the changelog entry.

Closes #326 · task `VortexStepMethod.jl-326`


CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Changed
+
+- Requires Julia 1.12 or 1.13; 1.10 and 1.11 keep resolving v5.1.1.
```
1-Bort-1, line 7: MINOR: "Requires Julia 1.12 or 1.13" is wrong: compat `"1.12, 1.13"` also allows 1.14 and later, and README/index.md say "1.12 or later". Users reading the changelog get a narrower range than the package actually accepts; "Requires Julia 1.12 or later" would match.

1-Bort-1 (commented):

**Good**

- Matches the card: `git diff --stat 23f129b HEAD` shows the same 10 files and +3149/−1478, with no changes the card leaves out
- Only the `test` job in CI.yml names versions (checked the file on 23f129b); the three `include:` entries correctly stay on 1.12
- Both manifests changed as small as claimed: the 1.12 `.default` moves only the package's own `version` 4.0.0→5.1.1, and the 1.13 one has `julia_version = "1.13.0"` and includes the `test`/`docs`/`examples` deps (Aqua, Documenter, CairoMakie)
- Removing the 1.11 sub-manifest copy branch is safe: `_sub_manifest` is used nowhere else in `bin/install`, and `MAIN_MANIFEST`/`MAIN_DEFAULT` still take their name from `_julia_major`
- `.gitignore` covers `Manifest*.toml`, so only the `.default` files are tracked, as §9 requires; the 1.11 `.default` is deleted rather than kept around (§2)
- Found-on-the-way items are recorded instead of fixed quietly (`juliaup default`, `juliaserver` tmux environment, the stale 1.11 docstring), and the macOS/Windows-on-1.12 risk is stated

**Not good**

- `CHANGELOG.md:7` — "Requires Julia 1.12 or 1.13" is wrong: compat `"1.12, 1.13"` also allows 1.14 and later, and README/index.md say "1.12 or later". Users reading the changelog get a narrower range than the package actually accepts; "Requires Julia 1.12 or later" would match.
- `julia = "1.12, 1.13"`: Julia compat is caret-style, so "1.12" already means [1.12, 2.0) and ", 1.13" adds nothing; harmless, but it reads as an upper limit it does not set
- `docs/src/functions.md:109` still links the Julia package docs under `pkgdocs.julialang.org/v1.11/`; it is outside the files this PR opens, so it belongs in the follow-up `cleanup:`
- The `CHANGELOG.md` `## Unreleased` header will conflict with #325, as the card says; whichever PR merges second has to fold its entry under the other's header, not add a second header

1-Bort-1:

Issue #318, opened 2026-09-14
correct_aoa defaults to false, so the solver runs as LL-3/4 and loses 24-59% of the induced drag


1-Bort-1:
`correct_aoa` defaults to `false` (`src/solver.jl:178`, `src/settings.jl:203`), so by default one velocity — the 3/4-chord control-point one — both reads the polars and turns the resulting force. Gaunaa, Li & Pirrung, *From thin airfoil theory to finite wings: consistent implementation of Lifting Line-based models*, TORQUE 2026, J. Phys. Conf. Ser. 3224 042030 (`OpenSourceAWE/Internal` `Papers/Gaunaa_2026.pdf`) calls that implementation **LL-3/4** and measures what it costs.

The rule is **TAT3**, boxed, p. 4:

> The flow direction that determines the lift and drag force directions in 2D UTAT is the undisturbed relative flow direction evaluated at the quarter-chord point.

with **TAT2** (p. 4) putting the *magnitude* at the 3/4-chord point, so the two points do different jobs. Footnote 2, p. 4, on using one point for both:

> Within the LL framework, this error vanishes only for straight wings with elliptic planform and very high aspect ratios.

### What it costs

Against a VLM reference at α = 10°, inviscid polars (`C_l = 2πα`, `C_d = 0`), 120 spanwise sections. `e_Osw,Blade` is the Oswald efficiency from the on-blade forces — i.e. induced drag:

| case | LL-3/4 `C_L` | LL-3/4 `e_Osw,Blade` |
|---|---|---|
| straight elliptic, AR 20 | −0.46 % | **−23.64 %** |
| straight elliptic, AR 4 | −6.72 % | **−58.81 %** |
| swept + anhedral elliptic (10 % sweep, 25 % anhedral) | +7.02 % | **−8.73 %** |

(Table 1 p. 8, Table 2 p. 9.) Lift survives; induced drag does not. p. 7:

> e_{Osw,Blade} is significantly underestimated by LL-3/4 (-59%) and LL-Ranneberg (-38%), whereas LL-1/4 and LL-Gaunaa, which both use the local conditions at the 1/4 chord point for the local force directions, remain within 1% of the VLM reference.

A kite is the low-aspect-ratio, curved case, not the straight high-AR one the footnote exempts.

### The machinery is already here

[...]

Issue #317, opened 2026-09-14
alpha is built from the full relative velocity, not its component perpendicular to the quarter-chord line


1-Bort-1:
`panel_inflow` forms the polar-lookup angle from the full relative velocity:

```julia
v_eff = 0.5 .* (va_1 .+ va_2) .+ v_ind
alpha = atan(dot(v_eff, z_airf), dot(v_eff, x_airf))
```

(`src/panel_aerodynamics.jl:119-120`, and open-coded on the hot path at `src/solver.jl:749-759`.)

Gaunaa, Li & Pirrung, *From thin airfoil theory to finite wings: consistent implementation of Lifting Line-based models*, TORQUE 2026, J. Phys. Conf. Ser. 3224 042030 (`OpenSourceAWE/Internal` `Papers/Gaunaa_2026.pdf`) states the rule as **CP1**, boxed, p. 3:

> Only the relative velocity component perpendicular to the local span direction, here defined as the quarter-chord bound vortex line, should be used in the inner 2D part of the LL framework.

and writes it as Eq. (1), p. 4, applied **before** the angle is formed:

> **V**^⊥_{3D,i} = **V**_{3D,i} − (**V**_{3D,i} · **e**_{s,i}) **e**_{s,i}

> α_{3/4,i} = atan2(**V**_{inner,3/4,i} · **e**_{n,i}, **V**_{inner,3/4,i} · **e**_{c,i})   (Eq. 3, p. 5)

The projection is missing here. It matters only in the denominator: `z_airf` is already perpendicular to the quarter-chord line, so `dot(v, z_airf) == dot(v_perp, z_airf)` exactly, but `x_airf` is the raw chord and carries a spanwise component on any swept panel.

Half of the rule is already obeyed elsewhere in the same function. The dynamic pressure is built from `v_span = cross(v_eff, y_airf)` (`src/panel_aerodynamics.jl:121`, `v_a_dist` at `src/solver.jl:731-766`), which is the span-perpendicular speed — simple sweep theory's `cos²Λ` about the quarter-chord line. So `q` is measured in the crossflow plane and `alpha` is not.

### Size

On the swept, twisted panel `test/panel/test_panel.jl` already builds, where the chord sits 45.128° out of the span-perpendicular plane (`cos(sweep) = 0.705526`):

| `v_eff` | `alpha` today | Eq. (1) then Eq. (3) |
|---|---|---|
[...]

PR #334, opened 2026-09-16
Test the solver against the references in the Python verification cases

### LINKED ISSUE #93: Add test cases from the Python verification cases
See: https://github.com/ocayon/Vortex-Step-Method/tree/main/tests/verification_cases

1-Bort-1:
### TL;DR

New `test/verification/test_verification.jl` ports the five cases in the Python `tests/verification_cases` that check against something outside the code: an elliptic wing against lifting-line theory, a curved Clark Y wing against RANS, a rectangular AR 12 wing against CFD, the V3 kite against RANS, and three horseshoe vortices against Biot-Savart. Until now nothing in the Julia suite compared a whole-wing CL or CD with theory or measured data. The existing reference tests only compare against the old thesis port, which is a copy of the same maths.

### What was ported and what was not

Each case keeps the Python geometry, flow speed, relaxation factor, `core_radius_fraction = 1e-20` and angles of attack, and every tolerance is copied unchanged from ocayon/Vortex-Step-Method@c82d521. The Python asserts comparing its "old" thesis code with its "new" package are left out, because they compare two Python codes. Geometry reuses the generators already in `test/utils.jl`. The V3 kite's 10 CAD section pairs are written into the test and split 4× with `SPLIT_PROVIDED` (37 sections, as in Python). The six CSVs are in `test/verification/data/`, with a README giving the source of each.

Julia passes every tolerance. The tightest margins are the V3 kite's CL at 9° (0.079 against 0.1) and the elliptic wing's VSM CD (0.00137 against 0.002):

| case | reference | Julia VSM | Julia LLT | tolerance |
| --- | --- | --- | --- | --- |
| elliptic AR 20, 3°/9° | CL 0.2991/0.8972, CDi 0.001424/0.01281 | CL 0.2954/0.8861, CD 0.001578/0.01418 | CL 0.2991/0.8967, CD 0.001423/0.01278 | LLT CL 1e-2, CD 1e-3 · VSM CL 1e-1, CD 2e-3 |
| curved Clark Y, 9° | CL 0.7921, CD 0.07507 | CL 0.7904, CD 0.1034 | CL 0.9093, CD 0.1029 | VSM 1e-1 · LLT−VSM CL 2e-1, CD 4e-2 |
| rectangular AR 12, 3°/6°/9° | CL 0.6094/0.8821/1.144 | CL 0.6252/0.8618/1.101 | CL 0.6374/0.8868/1.141 | VSM CL 1e-1 · LLT−VSM CL 2e-1, CD 4e-2 |
| V3 kite, 3°/6°/9° | CL 0.4608/0.6728/0.8605, CD 0.04815/0.06400/0.08148 | CL 0.4470/0.7020/0.9391, CD 0.0685/0.0960/0.1346 | – | VSM 1e-1 |

### Where the port differs from the Python test

- **Horseshoes:** the Python test sets the bound-vortex velocity to zero whenever the evaluation point has the filament's x (`if evaluation_point[0] == filament.x1[0]`), so it only ever tested the trailing legs. The Julia test adds all three filaments. The bound filament really does induce exactly zero here, because the evaluation point lies on its line, so the reference values are unchanged.
- **"Swept" wing:** Python's `swept_wing` case has sweep 0. It is named for what it is here: a rectangular AR 12 wing.
- **Clark Y polar:** Python's `clarky_maneia.csv` repeats its 20–32° block three times, which makes alpha non-monotonic, and its comment says radians although the values are degrees. Only the first block is kept.
- **Kite CD reference:** Python first resamples the CD data onto the CL data's angles and then interpolates again. Here CD is interpolated directly, which moves the reference by at most 3e-4 against a 0.1 tolerance.

### Found on the way

- `.github/workflows/CI.yml:22` has `fail-fast: false`, which the agent rules want gone. I did not change it here, since it is a separate `cleanup:` change.
- The drag tolerances are loose because upstream widened them in ocayon/Vortex-Step-Method@fac126b, when its VSM stopped moving alpha from 3/4 chord to 1/4 chord. Julia's VSM CD is 0.03–0.05 above the RANS values on the curved wing and on the kite at 6–9°. That is the drag #317 and #318 are about, and these tests will still pass after either fix.

### Verification

- [x] Reproduced first: n/a, these are tests of existing behaviour with no bug to reproduce
- [x] The tests can fail: making the semi-infinite trailing vortex 10 % stronger in `src/filament.jl` gave `18 passed, 4 failed` (elliptic VSM CD and the three semi-infinite horseshoe sums). With `src/` restored it went back to 22/22 green
- [x] `test/verification/test_verification.jl`: 22/22 pass. Through `runtests.jl` together with `body_aerodynamics/test_body_aerodynamics.jl`, which includes the same `utils.jl`: 4880/4880 pass, verification set 18.7 s cold (juliaserver)
- [ ] Local full suite (`agent ci-local`, Julia 1.12): still running when this was written · GitHub CI: pending
- Docs: n/a, no public symbols · Changelog: none, test-only · up to date with main
- Risk: the kite CL at 9° has 0.021 of headroom, so a physics fix that raises kite lift may need this reference looked at again, not the tolerance loosened.

### Scope

+350 / −0 across 10 files, all under `test/`. That is 226 lines of test (50 of them the kite's CAD coordinates), 107 lines of reference CSV, a 13-line data README, and one line each in `runtests.jl` and `test/README.md`.

Closes #93 · task `VortexStepMethod.jl-93`


test/verification/test_verification.jl
```diff
@@ -0,0 +1,226 @@
...
+    (deg2rad.(alpha), collect(cl), collect(cd), collect(cm))
+
+"""
+    wing_from_coordinates(coordinates, aero_model, aero_data=nothing;
+                          n_panels, spanwise_distribution=UNCHANGED)
+
+Refined wing with one section per leading/trailing-edge row pair of the 2N×3
+`coordinates`, which run from -y to +y.
+"""
+function wing_from_coordinates(coordinates, aero_model, aero_data=nothing;
```
1-Bort-1, line 39: MINOR: `wing_from_coordinates` is the same flip → `Wing(n; UNCHANGED)` → `add_section!` pairs → `refine!` loop written out twice in `test/body_aerodynamics/test_body_aerodynamics.jl` (lines ~32 and ~172), which includes the same `utils.jl`. Adding it only here makes a third copy; it belongs in `test/utils.jl`, with the body-aero copies switched to it now or in a `cleanup:` PR.

test/verification/test_verification.jl
```diff
@@ -0,0 +1,226 @@
...
+            (gamma=2.0, left=[0.0, -1.0, 0.0], right=[0.0, -3.0, 0.0],
+             velocity=[0.0, 0.0, 0.1061]),
+            (gamma=10.0, left=[0.0, 1.0, 0.0], right=[0.0, -1.0, 0.0],
+             velocity=[0.0, 0.0, -1.5915]),
+            (gamma=5.0, left=[0.0, 3.0, 0.0], right=[0.0, 1.0, 0.0],
+             velocity=[0.0, 0.0, 0.2653]),
+        )
+        for (gamma, left, right, velocity) in horseshoes
+            bound_velocity = zeros(3)
+            velocity_3D_bound_vortex!(bound_velocity, bound_filament(right, left),
```
1-Bort-1, line 201: MINOR: In all three horseshoes the evaluation point lies on the bound filament's line, so `bound_velocity` is exactly zero and the change sold as 'adds all three filaments' checks nothing Python did not. Either assert `bound_velocity == zeros(3)` so the claim is tested, or drop the call and say in the card that the case only covers the trailing legs.

1-Bort-1 (commented):

**Good**

- This is what #93 asked for: the five Python cases that check against theory, CFD, RANS or Biot-Savart, and no Python-vs-Python checks; I compared the diff with the card and every testset maps to a card row
- Geometry comes from the existing `generate_coordinates_*` and `flip_created_coord_in_pairs` in `test/utils.jl`; the new helper names (`read_columns`, `polar_vectors`, `max_error`, …) collide with nothing else in `test/`, which I checked with grep
- The card's margins match its table: kite CL at 9° is 0.9391−0.8605=0.079 and elliptic VSM CD is 0.01418−0.01281=0.00137
- The card shows the tests can fail: a 10 % stronger semi-infinite vortex turned 4 of the 22 checks red
- Reusing `left_velocity`/`right_velocity` in the finite-trailing check is safe: `velocity_3D_trailing_vortex!` overwrites `vel` in both branches (`src/filament.jl`)
- No line in `test_verification.jl` is over 92 characters (checked by grep); the `DelimitedFiles` and `Interpolations` dependencies are already in `test/Project.toml`
- The card explains each departure from Python (the sweep-0 name, the tripled Clark Y block, the direct CD interpolation) with its size against the tolerance

**Not good**

- `test/verification/test_verification.jl:39` — `wing_from_coordinates` is the same flip → `Wing(n; UNCHANGED)` → `add_section!` pairs → `refine!` loop written out twice in `test/body_aerodynamics/test_body_aerodynamics.jl` (lines ~32 and ~172), which includes the same `utils.jl`. Adding it only here makes a third copy; it belongs in `test/utils.jl`, with the body-aero copies switched to it now or in a `cleanup:` PR.
- `test/verification/test_verification.jl:201` — In all three horseshoes the evaluation point lies on the bound filament's line, so `bound_velocity` is exactly zero and the change sold as 'adds all three filaments' checks nothing Python did not. Either assert `bound_velocity == zeros(3)` so the claim is tested, or drop the call and say in the card that the case only covers the trailing legs.
- The card does not say what was searched for before the new helpers were added (§2), and the duplication below suggests the search missed test_body_aerodynamics.jl
- The local full suite and GitHub CI were still unticked when the card was written, so the only green result is from the verification file plus body_aerodynamics
- Kite CL at 9° has 0.021 of headroom; the card rightly says to revisit the reference rather than widen the tolerance, and that should hold when #317/#318 land
- `max_error` is the only helper without a docstring; its name is clear, but the file is otherwise consistent about docstrings

Issue #337, opened 2026-09-16
One name per apparent-wind quantity: `va`, `v_a`, `_va` and `va_*`...


1-Bort-1:
One name per apparent-wind quantity: `va`, `v_a`, `_va` and `va_*` made consistent

Third of the three parts of #147 (Refactoring part III). Stack it on #336 (`*_array` → vector/matrix names).

The convention agreed on #147: ufechner7 proposed `va` for the scalar norm, `va_vec` for the vector, and `va_dist` for the per-panel norms, with `_vec` as the suffix for any 1D vector. Bart accepted it as long as it holds everywhere, and on 2026-09-16 asked for consistent, readable naming across the package.

**Ask Bart before writing code.** The convention collides with public API that already uses `va` for the **vector**:
- `set_va!(body, va)`
- `BodyAerodynamics(...; va=)`
- `reinit!(...; va=)`
- the `Panel.va` field
- the `BodyAerodynamics.va` property

The plotting keywords are `v_a=` for the scalar: `plot_polars`, `plot_combined_analysis` and `generate_polar_data`. Linearize has `va_idxs=`.

Following the convention literally is a BREAKING rename of those keywords and that field. The alternative keeps the public names and applies the convention to everything else. Those two produce materially different code.

Survey of main at 2b312ea:
- 54 identifiers and 674 occurrences.
- Fields mix `_va`, `has_distributed_va`, `v_a_dist` (1D norms, on `BodyAerodynamics` and `LoopResult`), `VSMSolution._va_dist` (P×3 matrix), `va_unrefined_dist`, and `va_norm_dist`/`va_unit_dist` on `BaseResult`.
- Locals include `va_norm`, `va_unit`, `va_ref_vector`, `va_ref_mag`, `va_panel_mag`, `va_input`, `va_distribution`, `va1..3`, `dva_1/2`, and `delta_va` (a positional argument of the exported `section_pitch_rate`).
- Examples, tests and docs also call the same quantity `vel_app` (21), `v_app` (9, including settings docs) and `Uinf` (the thesis test port).
- Unrelated: `va`/`vb` in src/obj_adapter/obj_slice.jl:253-260 mean vertex a/b.

1-Bort-1:
[...]

Issue #336, opened 2026-09-16
Rename the `*_array` locals to say whether they are a vector or a...


1-Bort-1:
Rename the `*_array` locals to say whether they are a vector or a matrix

Second of the three parts of #147 (Refactoring part III). Stack it on the pull request from branch `agent/147-refactoring-part-iii` (typed empty containers). The `va` naming is the third part and comes after this one.

#147 asks to replace every `array` in a name with `vector` (1D) or `matrix` (2D). ufechner7 suggested the suffix `_vec` for a 1D vector on that thread, and Bart agreed as long as the naming is consistent everywhere. A survey of main at 2b312ea found 22 identifiers and 241 occurrences, all of them locals, positional arguments, or the private `calc_norm_array!`. None is exported, a struct field, or a keyword, so no public API moves.

- 2D (P×3): `va_array`, `va_unit_array`, `x_airf_array`, `y_airf_array`, `z_airf_array`, `relative_velocity_array`, `v_acrossz_array`
- 1D: `va_norm_array`, `chord_array`, `v_normal_array`, `v_tangential_array`, `va_magw_array`, `mu_array`, `cl_array`, `cd_array`, `cm_array`, `panel_width_array`, `force_array`, `moment_array` (length 3), `v_a_array`, `alpha_array`
- Files: src/solver.jl, src/body_aerodynamics.jl, test/bench.jl, test/body_aerodynamics/test_body_aerodynamics.jl, test/solver/test_solver.jl

The struct fields holding the same data already use `_dist` (`VSMSolution._x_airf_dist`, `_va_dist`, `_chord_dist`). Decide whether the locals follow the fields or the suffix. A per-panel quantity reads as `_dist` whether it is 1D or 2D, and that decision belongs in the PR card. There are no 3D names; `AIC` and `SectionAero.cp`/`cf` are `Array{T,3}` fields and keep their names.

Also noticed: `test/body_aerodynamics/test_body_aerodynamics.jl:66-67` and `:101-102` size `va_norm_array`/`va_unit_array` with `length(coord)` on a matrix, which counts elements rather than rows.

1-Bort-1:
[...]

Issue #147, opened 2025-03-27
Refactoring part III


ufechner7:
- [x] do not use untyped vectors like  `vertices = []`
- [x] more consistant naming for v_a: In the moment we have `_va`, va, `v_a` and `v_a_array` 
- [x] replace all occurances of `array` with either `vector` (1D) or `matrix` (2D). Not sure if we are using 3D arrays, I don't think so.

ufechner7:
> I like `va`, as it looks better in `va_array`, `_va`, etc.

The point is, they have different meanings, so we will need different names:
1. norm of the apparent wind speed (scalar)
2. apparent wind speed vector
3. vector of apparent wind speeds per panel

In the moment I don't know if 3. is a vector of scalars or a vector of vectors.

1-Bart-1:
We have talked about this before on telegram, as long as the naming is consistent everywhere these names are fine.

1-Bart-1:
@1-Bort-1 implement consistent and readable naming in this package.

1-Bort-1:
This branch does the first box only: every untyped empty container (`vertices = []`, `plots = []`, ...) gets its element type, in src, the Makie extension, tests and an example. The two renames are stacked follow-ups: #336 (`*_array` → vector/matrix) and then #337 (one `va` convention, which starts by asking about the public `va=`/`v_a=` keywords).

1-Bort-1:
First part (typed empty containers) is e971c74 on agent/147-refactoring-part-iii, whose PR opens with this turn. The two renames are queued as #336 (`*_array`) and #337 (`va`).

PR #338, opened 2026-09-16
Give every untyped empty container its element type

1-Bort-1:
### TL;DR

`vertices = []`, `plots = []`, `filaments_plot = []` and the other bare `[]`/`Any[]` literals now say what they hold: in src, the Makie extension, tests and the stall example. This is the first box of #147. Typing the list in the Makie extension surfaced a real inconsistency: `plot!(ax, body)` drawn as flat panels returned one list per panel, while every other branch returned a flat list. Now all branches return one flat `Vector{Makie.AbstractPlot}`.

### What changed

- `read_faces` builds `Vector{Vector{Float64}}` / `Vector{Vector{Int64}}`, and `calculate_filaments_for_plotting` builds the `Vector{Tuple{Vector{Float64}, Vector{Float64}, String}}` its docstring already promised. Both used to return `Vector{Any}`.
- Makie extension: both `plot!` methods collect into `Makie.AbstractPlot[]`. The static body path now uses `append!` instead of `push!`, because pushing a panel's plot vector into a typed list throws. That is the one behaviour change and it has a changelog line. `plot_polars` types its `polar_data_list` and `cm_data_list`.
- `VSMSettings.wings` defaults to `WingSettings[]` instead of a `[]` that kwdef converted anyway.
- Tests: the Python-port helpers in `test/thesis_oriol_cayon.jl` get `Dict{String, Any}[]`, `Vector{Float64}[]` and so on. Three `push!` loops became comprehensions, and `plot_polars(Any[], Any[], …)` became `plot_polars(Solver[], BodyAerodynamics[], …)`.
- `examples/stall_model.jl` adds each rib to the wing as it reads the CSV row, instead of filling an untyped `rib_list` first.

Searched for `= []`, `Any[]` and bare `[]` literals across src, ext, test, examples, docs/src, scripts and README. None are left.

### Left as they are

- `Vector{Any}[]` rows in `geometry_gen.jl`, `obj_to_yaml.jl` and `SurfplanAdapter.jl`. These are YAML table rows that mix an `Int` id with `Float64` coordinates. Typing them as `Float64` would print the id as `3.0`.
- `Tuple[]`, `NamedTuple[]` and `Ref{Any}`: these are typed, only abstractly.
- The extension's global observable stores are `Dict()`. `PANEL_MESH_OBSERVABLES` is keyed by `(body_id, panel_idx)` in two places and by a bare panel `objectid` in the third (`plot(panel)`). The mixed key is the thing to fix there, and it is a different concern from typed vectors.

### The rest of #147

The two renames touch about 900 names, and `va=`/`v_a=` are public keywords, so they are queued as a stack on top of this PR:
1. #336: `*_array` locals to vector/matrix names (no public API)
2. #337: one `va` convention. It starts by asking whether the public `va=` keyword, `set_va!` and `Panel.va` follow the convention agreed on #147.

Not stacked on #325, which also edits `obj_geometry.jl` and `test_kite_geometry.jl`, but in hunks that don't overlap these.

### Verification

- [x] Reproduced first: the new checks on main's code evaluate `Any[[0.0, 0.0, 0.0], …] isa Vector{Vector{Float64}}`, `Any[(…, "magenta"), …] isa Vector{Tuple{…}}`, `Any[Any[Mesh{…}, …]] isa Vector{Makie.AbstractPlot}` and `20 == 40` (nested per-panel list)
- [x] `test/panel/test_panel.jl`, `test/ram_geometry/test_kite_geometry.jl`: 3 failed before, green after. `test/plotting/test_plotting.jl`: 3 failed before, green after.
- [x] `runtests.jl` filtered to panel, kite geometry, settings, obj adapter, body aerodynamics and plotting: 5085 passed, 1 broken (the existing `@test_broken` in Kite Geometry). `yaml_geometry` (includes wing deformation): 178 passed. `examples/stall_model.jl` runs to the end.
- These ran in a bare `julia +1.12 --project=test`, not juliaserver: the box's REPL sweeper stopped this task's session every minute (1-Bart-1/Agents#413)
- [x] Up to date with main (2b312ea) · no new or renamed public symbol, docs build not run · no REUSE in this repo
- Local CI mirror: started, not finished at the time of writing · GitHub CI: not run yet, starts when this PR opens
- Risk: a downstream caller that indexed `plot!(ax, body)` as `plots[i]` per panel now gets individual plots.

### Scope

+60 / −52 across 11 files. The growth is the three new type checks and the changelog line; the rest is one-for-one replacements.

Refs #147 · task `VortexStepMethod.jl-147`


examples/stall_model.jl
```diff
@@ -32,19 +32,15 @@ csv_file_path = joinpath(
 )
 
 df = CSV.read(csv_file_path, DataFrame)
-rib_list = []
-for row in eachrow(df)
-    LE = [row.LE_x, row.LE_y, row.LE_z]
-    TE = [row.TE_x, row.TE_y, row.TE_z]
-    push!(rib_list, (LE, TE, LEI_AIRFOIL_BREUKELS,
-                     lei_poly_coeffs(row.d_tube, row.camber)))
-end
```
1-Bort-1, line 35: MINOR: With `rib_list` gone, a stray blank line now separates `df = CSV.read(...)` from its only use, and the `# Create wing geometry` comment now also sits over the CSV read loop. Moving the read down next to the loop would keep reading the data and building the wing in one place.

1-Bort-1 (commented):

**Good**

- The commit's own diff (HEAD~1..HEAD) is 11 files, +60/−52, and matches the card's file list and scope exactly
- A search of src, ext, test, examples and docs for `= []`, `Any[]`, `([]` and `, []` finds nothing left, which backs the card's claim
- The static-body change from `push!` to `append!` is correct: the Panel `plot!` returns its own list (mesh + border), so the old result was nested and the new one is flat like the observable and airfoil branches; there is a changelog line and a test (`length == 2 * panels`)
- The declared element types match what gets pushed: `generate_polar_data` builds `polar_data` from vectors plus distributions, and the thesis helpers' `ringvec`/`wingpanels` dicts hold only `Vector{Float64}` values
- `calculate_filaments_for_plotting` now returns the tuple type its docstring already promised, and a new unit test pins it rather than a slow integration test
- The `plot_polars(Solver[], BodyAerodynamics[], …)` test calls still fit the untyped `solver_list`/`body_aero_list` signature, so no API changed
- Deferring the `va`/`*_array` renames to #336 and #337 keeps this PR to one idea, as §1 asks

**Not good**

- `examples/stall_model.jl:35` — With `rib_list` gone, a stray blank line now separates `df = CSV.read(...)` from its only use, and the `# Create wing geometry` comment now also sits over the CSV read loop. Moving the read down next to the loop would keep reading the data and building the wing in one place.
- The mixed `PANEL_MESH_OBSERVABLES` key (`(body_id, panel_idx)` vs. a bare panel `objectid`) is a real inconsistency found on the way, but it is only described under 'Left as they are' with no linked issue, which §5 asks for
- `NamedTuple{(:cmx, :cmy, :cmz), NTuple{3, Vector{Float64}}}[]` at ext line 934 is a long inline type; it is correct, but a reader has to parse it to see it just means 'cm columns'
- Line 282 of `thesis_oriol_cayon.jl` repeats the `Dict{String, Any}[]` literal from line 127, so the two resets have to be kept in step by hand
- The only checks listed are local and filtered, and the local CI mirror had not finished; the full suite and GitHub CI are still unconfirmed

1-Bart-1 (approved):

Issue #345, opened 2026-09-16
stability_derivatives gives roll, pitch and yaw rate derivatives...


1-Bort-1:
stability_derivatives gives roll, pitch and yaw rate derivatives about the reference point

Second part of #330. The first part (branch `agent/330-add-rigid-body-stability-derivatives-and`, pull request from #330) adds `stability_derivatives` with the angle of attack and sideslip derivatives and `trim_angle`; stack this on that pull request.

Blocked on #329: `set_va!` takes the rotation about the origin, while `linearize`'s moments are about `solver.reference_point`, so the `omega` columns of `linearize` are only the rate derivatives once #329 lets the rotation be taken about the reference point and `linearize` passes `solver.reference_point` through. Build on #329's pull request, not on a local copy of `omega × (r - r0)`.

What to add: `dC/dp`, `dC/dq`, `dC/dr` for C in CFx..CMz from the `omega_idxs` columns, in p̂ = pb/2V, q̂ = qc/2V, r̂ = rb/2V (Python `compute_rigid_body_stability_derivatives`, ocayon/Vortex-Step-Method, uses c = S/b for q̂ while its moment coefficients use the max chord; this package's moment coefficients use `body_aero.c_ref`, decide which and say so). Test against central differences of `solve!` at a reference point away from the origin.

1-Bort-1:
Adding dC/dp̂, dC/dq̂, dC/dr̂ for CFx..CMz to stability_derivatives from linearize's omega columns, turning the body about solver.reference_point. Stacked on #346, with #353 (#329's reference_point on set_va!) merged in, so it stays draft until #353 merges. q̂ uses body_aero.c_ref, the chord the moment coefficients use.

Issue #344, opened 2026-09-16
linearize and make_dual_shadow accept a BodyAerodynamics with more...


1-Bort-1:
linearize and make_dual_shadow accept a BodyAerodynamics with more than one wing

Split out of #330 (stability derivatives and trim search on linearize), which needs it to cover multi-wing bodies.

`linearize` (`src/solver.jl`, the `length(body_aero.wings) == 1` guard) and `make_dual_shadow` (its own one-wing guard) reject a body with two wings. The `va` and `omega` columns need nothing per wing; `theta_idxs`/`delta_idxs` are one entry per unrefined section of the one wing, so for several wings they would run over the unrefined sections of all wings in order, matching `moment_unrefined_dist`, which the solver already sizes by the sum over wings.

Test: the two-wing bodies in `test/body_aerodynamics/test_body_aerodynamics.jl` (`solve! on a two-wing body`), Jacobian against AutoFiniteDiff as `test/solver/test_forwarddiff.jl` does for one wing.

1-Bort-1:
Merged main (#352), CHANGELOG only. Now addressing the review on #355: a test that the twist and deflection split hands each wing its own angles, tighter bounds on the far-apart test, and one helper for each wing's unrefined-section range, used by calc_forces! and unrefined_deform!. Opening issues for ELLIPTIC on several wings and for wings[1].spanwise_direction.

Issue #330, opened 2026-09-16
Add rigid-body stability derivatives and a trim-angle search, as in the Python package


1-Bort-1:
The Python package has rigid-body stability derivatives and a trim-angle search; this package has neither. The closest thing here is `linearize` (`src/solver.jl:1266`), which returns the dimensional Jacobian of forces and moments with respect to `va` and `omega` (about the origin), for one wing only.

What Python has, in `src/VSM/stability_derivatives.py` and `src/VSM/trim_angle.py` (ocayon/Vortex-Step-Method@a397bcd, v2.0.0):

- `compute_rigid_body_stability_derivatives(body_aero, solver, angle_of_attack, side_slip, velocity_magnitude, roll_rate, pitch_rate, yaw_rate, step_sizes, reference_point, nondimensionalize_rates=True)`: `dC{x,y,z}_d{alpha,beta,p,q,r}` and `dCM{x,y,z}_d{...}` by central differences, rate derivatives optionally in p̂ = pb/2V, q̂ = qc/2V, r̂ = rb/2V.
- `map_derivatives_to_aircraft_frame`: VSM frame (x back, z up) to aircraft frame (x forward, z down).
- `build_malz_coeff_table_from_solver`: the derivatives as quadratics in alpha, the table format of Malz et al.'s kite models.
- `compute_trim_angle(body_aero, solver, side_slip, velocity_magnitude, ..., alpha_min, alpha_max, reference_point)`: coarse alpha sweep for a CMy sign change, bisection, and `dCMy/dalpha < 0` as the stability check.

Proposal: build both on `linearize` rather than on finite differences — the alpha/beta derivatives are the chain rule through `va`, and p/q/r already come out of the `omega` columns once the rotation is taken about the reference point (#329). Needs support for more than one wing in `linearize` to cover multi-wing bodies.

Upstream caveats worth not porting: `build_malz_coeff_table_from_solver` on Python main still passes the removed `yaw_rate/pitch_rate/roll_rate` keywords to `va_initialize`, and `_evaluate_with_angles` drops `body_axis` (fixed only on the `improve_plotting` branch).

Found while surveying the Python package for #113.

1-Bort-1:
[...]

Issue #329, opened 2026-09-16
set_va! rotates the body about the origin only, and condition.yaw_rate is read but never applied


1-Bort-1:
`set_va!` rotates the body about the origin and nothing else, and the YAML `condition.yaw_rate` is read into `ConditionSettings` and never used.

**Reference point.** `set_va!(body_aero, va, omega)` (`src/body_aerodynamics.jl:1091`) sets each panel's inflow to `va - omega × control_point` (`:1106`). The Python package takes the rotation about a reference point, `va_i -= omega × (cp_i - r0)` (`src/VSM/core/BodyAerodynamics.py:559`, keyword `reference_point`, ocayon/Vortex-Step-Method@b5cd5f1). Here the only way to move the pivot is `kite_body_origin`, which shifts the whole geometry. The moment reference point (`Solver.reference_point`) and the rotation centre are therefore independent inputs in Python and can only be matched here by moving the mesh.

**`yaw_rate` ignored.** `ConditionSettings.yaw_rate` (`src/settings.jl:15`, `[°/s]`) is in every shipped `vsm_settings*.yaml`, but `set_va!(body_aero, settings)` (`src/body_aerodynamics.jl:1186-1197`) reads only `alpha`, `beta` and `wind_speed` and calls `set_va!(body_aero, va)` with no `omega`. A non-zero `yaw_rate` in a settings file has no effect; `examples/V3_kite.jl` and `examples/pyramid_model.jl` only print it in plot titles.

Proposal: a `reference_point` keyword on `set_va!` (default the origin, so results do not move), and either pass `yaw_rate` through as `omega` about the body z axis or remove the field. The per-panel `set_va!(body_aero, va_distribution::AbstractMatrix; pitch_rate_dist)` method would take the same keyword.

Found while surveying the Python package for #113.

1-Bort-1:
Adding a `reference_point` keyword to `set_va!(body_aero, va, omega)` (inflow `va - omega × (cp - reference_point)`, default the origin), stored on `BodyAerodynamics` so `reinit!`, the `omega` setter, linearize and the plot helpers keep the pivot. `set_va!(body_aero, settings)` passes `condition.yaw_rate` through as omega about body z. Nothing open touches this code. Leaving out the per-panel matrix method, which takes no omega and so has no pivot.

PR #346, opened 2026-09-16
Add stability_derivatives (angle of attack, sideslip) and trim_angle, built on linearize

1-Bort-1:
### TL;DR

New `stability_derivatives(solver, body_aero, alpha, beta, wind_speed)` returns the six force and moment coefficients and their derivatives with respect to angle of attack and sideslip. New `trim_angle(solver, body_aero, beta, wind_speed)` finds where `CMy` changes sign and returns the slope `dCMy/dalpha` at each trim, so you can see whether it is stable. The Python package has both and this package had neither. Both are built on `linearize`, as #330 proposes, rather than on a second finite-difference path.

### How it works

- **Derivatives.** `linearize(...; va_idxs=1:3, aero_coeffs=true)` gives the coefficient Jacobian with respect to `va`. That is multiplied by `dva/dalpha` and `dva/dbeta`, which ForwardDiff takes from `apparent_wind`. Moments are about `solver.reference_point`, the rotation rate is whatever `body_aero.omega` holds, and extra keywords go to `linearize` (e.g. `backend=nothing` for a `NONLIN` solver, since ForwardDiff only runs on `LOOP`).
- **Inflow formula.** `apparent_wind(alpha, beta, wind_speed)` is the formula `set_va!(body_aero, settings)` already used, `V·[cosα cosβ, sinβ, sinα cosβ]`, moved out so both callers share it. The old docstring labelled those components forward/right/down; the body frame is x back, z up, so the docstring now just names the formula.
- **Trim search.** It sweeps `alpha_range` (default −5° to 15° in 2° steps, in radians), bisects every bracket where `CMy < 0` flips to `alpha_tol` (default 1e-5 rad), and takes each slope from `stability_derivatives`. It returns every trim found, and an empty vector if there is none.
- **Trim failures.** Every solve behind a trim runs with `throw_on_fail`: the sweep, the bisection and the slope's `linearize`. A solve that misses the solver's tolerances throws `SolveFailure` instead of steering the bisection. On the test wing with `max_iterations=2`, the version before the review silently returned a trim at 1.00° with slope −1.56; converged, the answer is 1.666° and −1.11.
- **Trim keywords.** `trim_angle` takes `backend` explicitly rather than passing keywords through to `linearize`. With pass-through, a `reference_point=` reached the slope but not the `CMy` sweep, so the trim and its stability verdict could be about two different points. Both review comments are addressed in b50a2b5.

### Where this differs from the Python package

- **Units.** Angles are in radians, like the rest of this Julia API. Python takes degrees.
- **Sideslip.** The inflow keeps `cos β` on z, as `set_va!(settings)` has it. Python main uses `sin α` without it, so the numbers differ from Python at nonzero sideslip.
- **Trim results.** Python raises when it finds more than one stable trim, and falls back to "closest to zero" when it finds none. Here every sign change is returned, and none means an empty vector.
- **Aircraft frame and Malz table.** Not ported. The aircraft-frame mapping is six sign flips a caller can apply. The Malz table is broken on Python main (#330 lists why) and nothing here asked for it.

### Split

#330 holds more than one idea, so this is part 1:

- #345 adds the p/q/r rate derivatives. They need the rotation taken about the reference point (#329). Until then `linearize`'s `omega` columns rotate about the origin while moments are taken about `solver.reference_point`, so porting the rates now would mean a second copy of `ω × (r − r0)`.
- #344 lets `linearize` and `make_dual_shadow` take more than one wing. Until it lands, `stability_derivatives` throws on a multi-wing body, and so does `trim_angle` once it has found a trim.

### Found on the way (not changed)

- **Examples still write the inflow formula by hand.** `examples/billowing.jl` has the same formula. `examples/linearize_check.jl` leaves `cos β` off z. `apparent_wind` is private, so I left the examples alone.
- **CI does not stop at the first failure.** `.github/workflows/CI.yml` sets `fail-fast: false`. Changing that is its own `cleanup:` PR.
- **Before searching:** I searched `src/`, `test/`, `examples/` and `docs/` for stability, derivative, trim, malz and bisect. Nothing to reuse; the one bisection, in `obj_slice.jl`, carries its own per-step payload.

### Verification

- [x] **Reproduction:** n/a, new feature.
- [x] **`test/solver/test_stability.jl`, red before, green after.**
  - Before the implementation, `include` gave `0 passed, 0 failed, 1 errored` because `stability_derivatives`, `trim_angle` and `apparent_wind` were undefined.
  - After: 15/15 in juliaserver.
  - The new `a solve that misses the tolerances throws` testset was red on the code before b50a2b5: it returned a trim instead of throwing.
  - The new `a NONLIN solver with backend=nothing finds the same trim` testset covers the finite-difference path. It matches the `LOOP` + ForwardDiff trim to 1e-4.
  - On a rectangular POLAR_VECTORS wing, the derivatives agree with central differences of `solve!` (step 1e-4 rad) to about 1e-8 relative, e.g. `dCFz/dα` 4.414345384 against 4.414345369. Tested to rtol 1e-4.
  - With section `cm = 0.05` and moments about the leading edge, it finds one trim at 1.666° with slope −1.11 /rad. With `cm = −0.05` and moments about the trailing edge, one trim at −0.050° with slope +3.34 /rad.
- [x] **`test/body_aerodynamics/test_body_aerodynamics.jl`:** green, including `set_va! with VSMSettings` (5/5), which covers the moved formula.
- [x] **Merged with main, up to date at 83867a3.**
  - #352 (viscous drag correction): the only conflict was the two Added entries in `CHANGELOG.md`; both are kept. `test_viscous_drag_correction.jl` passed 90/90 on that merge.
  - #349 (`va` rename): the conflict was the body of `set_va!(body_aero, settings)`. It now calls `apparent_wind`, and the vector local is `va_vec`, as in `stability_derivatives`. `test_body_aerodynamics.jl` is green on the merged tree.
- [x] **Docs build:** clean at b50a2b5, no missing docstrings. No REUSE in this repo.
- [x] **Local CI mirror on b50a2b5** (Julia 1.12, fail-fast): PASS in 9 min, Aqua included. **GitHub CI on b50a2b5:** PASS (all six jobs and codecov/patch).
- `jetls` is not installed on the box, so it was not run.
- **Risk:** `dCFy/dβ` is exactly zero on a flat wing, so the side-force derivative is only checked against the finite difference where it is zero. The other five sideslip derivatives are nonzero and checked.

### Scope

+209 / −25 across 8 files:
- `src/stability.jl` (92 lines, about half of them docstrings)
- `test/solver/test_stability.jl` (92 lines)
- `src/body_aerodynamics.jl`: `apparent_wind` moved out of `set_va!(settings)`, and its docstring cut from 25 lines to 3, net −12
- one line each in `runtests.jl`, the export list and the include list
- docs and CHANGELOG entries

Stack: 1/3 → #345 (rate derivatives, blocked on #329), #344 (multi-wing `linearize`).

Refs #330 · task `VortexStepMethod.jl-330`


src/stability.jl
```diff
@@ -0,0 +1,81 @@
...
+        alpha_range=deg2rad.(-5:2:15), alpha_tol=1e-5, kwargs...)
+    is_nose_down =
+        alpha -> pitch_moment_coeff(solver, body_aero, alpha, beta, wind_speed) < 0
+    nose_down = is_nose_down.(alpha_range)
+    trims = @NamedTuple{alpha::Float64, dCMy_dalpha::Float64}[]
+    for i in 1:length(alpha_range)-1
+        nose_down[i] == nose_down[i+1] && continue
+        alpha = bisect_sign_change(is_nose_down, alpha_range[i], alpha_range[i+1],
+            alpha_tol)
+        derivatives = stability_derivatives(solver, body_aero, alpha, beta, wind_speed;
```
1-Bort-1, line 46: MINOR: `kwargs` reach `linearize` and so `solve!` for the slope, but not the `CMy` sweep in `pitch_moment_coeff`. A caller passing `reference_point=` (a `solve!` keyword) gets trims found about `solver.reference_point` but slopes about another point, so the stable/unstable verdict can be silently wrong.

src/stability.jl
```diff
@@ -0,0 +1,81 @@
...
+
+"""
+    pitch_moment_coeff(solver, body_aero, alpha, beta, wind_speed)
+
+`CMy` of `body_aero` solved at angle of attack `alpha` [rad], sideslip `beta` [rad] and
+`wind_speed` [m/s], at the rotation rate `body_aero.omega`.
+"""
+function pitch_moment_coeff(solver, body_aero, alpha, beta, wind_speed)
+    set_va!(body_aero, apparent_wind(alpha, beta, wind_speed), body_aero.omega)
+    return solve!(solver, body_aero).moment_coeffs[2]
```
1-Bort-1: MINOR: `pitch_moment_coeff` never checks `solver.lr.converged`, and line 48 throws away `derivatives.converged`. A trim bisected on unconverged `CMy` values comes back looking just like a good one.

1-Bort-1 (commented):

**Good**

- The derivatives are built on `linearize`, as the card promises: `va_idxs=1:3` with `theta_idxs=nothing`, and `omega` read from `body_aero.omega` at `src/solver.jl:1319-1321`, so there is no second finite-difference path.
- The formula in `apparent_wind` matches the removed body of `set_va!(settings)` term for term, and that method now calls it, so the inflow has one source.
- `dva/dalpha` and `dva/dbeta` come from ForwardDiff over `apparent_wind`, so the formula's derivative is not written out a second time.
- The docstring fix fits the geometry: the test wing runs from the leading edge at x=0 to the trailing edge at x=1, and `docs/src/settings.md:54` has z = up, so the old forward/down labels were wrong.
- The tests name what they protect, check both slope signs and a range with no trim, and compare against central differences of `solve!` rather than against `linearize` itself.
- Public symbols are in `functions.md`, private helpers in `private_functions.md`, and CHANGELOG and export and include lines are added, matching the plan in the card.

**Not good**

- `src/stability.jl:46` — `kwargs` reach `linearize` and so `solve!` for the slope, but not the `CMy` sweep in `pitch_moment_coeff`. A caller passing `reference_point=` (a `solve!` keyword) gets trims found about `solver.reference_point` but slopes about another point, so the stable/unstable verdict can be silently wrong.
- `src/stability.jl:61` — `pitch_moment_coeff` never checks `solver.lr.converged`, and line 48 throws away `derivatives.converged`. A trim bisected on unconverged `CMy` values comes back looking just like a good one.
- `coeffs_at` in the test is `pitch_moment_coeff` returning all six coefficients; one private src helper returning all six, with `trim_angle` taking `[5]`, would serve both.
- The logic of the trim predicate sits in the closure `is_nose_down` (§6); a named helper, or `pitch_moment_coeff(...) < 0` passed straight in, reads the same without it.
- The trim tests use the default `Solver`, where `use_gamma_prev` is on, so each bisection step starts from the last step's gamma; the derivative test turns it off, and the trim tests do not say why they leave it on.
- The card says to pass `backend=nothing` for a `NONLIN` solver, but no test covers that path through `stability_derivatives` or `trim_angle`.
- `trims` is fixed to `Float64`, so a `Float32` or dual `alpha_range` is converted without warning; fine for now, but worth a line in the docstring if it is intended.

1-Bort-1:
**Local full suite: PASS** (9 min, Julia 1.13.0, one cell of the matrix)

src/stability.jl
1-Bort-1: Fixed in b50a2b5: trim_angle takes an explicit backend instead of splatting kwargs, so the sweep, the bisection and the slope all take moments about solver.reference_point.

src/stability.jl
1-Bort-1: Fixed in b50a2b5: the sweep, bisection and slope solves all run with throw_on_fail, so an unconverged solve throws SolveFailure instead of returning a trim (new testset, red on the old code: max_iterations=2 returned a trim at 1.00° with slope -1.56).

1-Bort-1:
**CI: `Julia 1.12 - windows-latest - x64 - pull_request` failed** on `0da60e375`.

https://github.com/OpenSourceAWE/VortexStepMethod.jl/actions/runs/35530963060/job/106131345658

1-Bart-1:
> Examples still write the inflow formula by hand

Fix this in this PR

1-Bort-1:
Fixed in c8264df: apparent_wind is exported, and the six examples, the README and the docs pages call it instead of writing out the formula.

1-Bart-1:
I would like to see some figs in the PR body to make it easier to review.

1-Bort-1:
Added three figures to the #346 description: the trim search on the V3 kite, the V3 derivatives against central differences, and the two test wings (collapsed).

1-Bart-1 (approved):

PR #273, opened 2026-09-05
Take the panel normal from the quarter-chord step

1-Bart-1:
Fixes #272.

`panel_axes` built the panel normal from the leading-edge step:

```julia
span_vec = panel_span_vector(le_1, te_1, le_2, te_2)   # quarter-chord step
y_airf   = orient .* (span_vec ./ width)
z_cross  = cross(x_airf, le_1 .- le_2)                 # leading-edge step
```

The bound vortex, `width` and `y_airf` all come from the quarter-chord step, so
`z_airf` was the odd one out. On a panel whose two sections have
differently-directed chords the normal is then not square to the vortex the loads
are built from, and since `alpha = atan(v·z_airf, v·x_airf)` feeds every polar
lookup, the error propagates into `cl`, `cd` and `cm` for that panel.

One character of the fix: `cross(x_airf, span_vec)`. The frame now closes as
`z_airf = x_airf × y_airf`, which is what the `Panel` field docstring
(`src/panel.jl:28`) has always claimed.

### Why it went unnoticed

Taper alone cannot trigger it. `(le_1 - le_2) - span_vec` equals
`¼(chord_vec_2 - chord_vec_1)`, which is purely chordwise when the two chords are
parallel, and `cross(x_airf, ·)` annihilates the chordwise part exactly. It takes
twist, sweep or dihedral — something that changes the chord *direction* between
the two sections — for the normals to separate at all.

The reference wings barely do that:

| reference wing | angle(LE step, QC step) | normal changes? |
|---|---|---|
| `:rectangular` | 0.84°, from its ±0.5 rad twist | yes |
| `:curved` | 0.00° — LE and TE share `(y, z)` | no |
| `:elliptical` | 34.01° | no — planar, so the normal is `±ẑ` regardless |

The elliptical wing separates the two steps by 34° and still cannot see the
difference. Nothing in the suite combines sweep with a changing chord direction,
which is the only regime where the choice matters.

### Changes to the reference oracle

`test/thesis_oriol_cayon.jl` makes the same mix (its `x_airf` is the normal, its
`z_airf` the spanwise axis):

```julia
x_airf = cross(VSMpoint - LLpoint, section["p2"] - section["p1"])  # LE step
z_airf = bound["x2"] - bound["x1"]                                 # QC step
```

so the inconsistency is inherited from the original formulation rather than
introduced by the port, and the oracle has to move with the code or it pins the
old convention. It now takes the normal from the bound filament it already builds
[...]

Issue #272, opened 2026-09-05
Panel normal z_airf is built from the leading-edge step, not the quarter-chord step the bound vortex uses


1-Bart-1:
## Summary

In `panel_axes` (`src/panel_aerodynamics.jl:58-69`) the panel's spanwise axis, its
width and its bound filament all come from the **quarter-chord** step, but the
normal comes from the **leading-edge** step:

```julia
span_vec = panel_span_vector(le_1, te_1, le_2, te_2)   # 0.75·le + 0.25·te
width    = smooth_norm(span_vec)
x_airf   = chord_vec ./ smooth_norm(chord_vec)
y_airf   = orient .* (span_vec ./ width)
z_cross  = cross(x_airf, le_1 .- le_2)                 # <-- leading edge, not quarter chord
z_airf   = orient .* (z_cross ./ smooth_norm(z_cross))
```

`panel.bound_point_1/2` and `filaments[1]` lie on the quarter-chord line, so
`z_airf` is not perpendicular to the vortex line the circulation sits on. The
effective angle of attack is taken against that normal
(`src/solver.jl:717-726`, `atan(v·z_airf, v·x_airf)`), so a rotation of `z_airf`
about `x_airf` feeds straight into every panel's alpha.

## Size of the discrepancy

Measured on a 44-panel ram-air kite (SK100, strongly raked tip), settled under
load. Second column is the angle between `z_airf` and `normalize(x_airf ×
(bound_point_2 − bound_point_1))`:

```
panel   LE step vs QC step   z_airf vs QC-plane normal   sweep(QC)
   1           6.97°                   5.58°               48.8°
   4           7.68°                  13.44°               57.2°
   5          10.26°                   7.21°               12.4°
  20           1.71°                   1.40°                1.5°
  41           7.83°                  13.82°              -57.2°
  44           6.98°                   5.71°              -48.7°
```

Near zero at the root, ~7° over most of the span, up to 13.8° on the outer
panels. It grows with taper, so it is largest exactly where the chord is
shortest and the panels are most sensitive.

## Suggested fix

Take both from the same span reference:

```julia
z_cross = cross(x_airf, -span_vec)
```

If the leading-edge step is deliberate — it tracks the physical LE, arguably the
better reference for a cambered section's plane — then the bound filament
arguably ought to follow it too. Either way, mixing the two references inside one
frame looks unintended.

## Related, possibly by design

Two things in the same area that only bite at high sweep, raised for the record
[...]

PR #350, opened 2026-09-16
Spread sliced sections evenly over the span, not over leading-edge arc length

### LINKED ISSUE #275: Section slicing spreads stations by leading-edge arc length, which collapses to nothing in span at a raked tip
## Summary

`station_indices` (`src/obj_adapter/obj_slice.jl:465`) spreads its section targets
over equal **leading-edge arc length**:

```julia
return [argmin(abs.(arclen .- t)) for t in range(inner + d, outer - d, n)]
```

Near a closed or raked tip the leading edge runs almost straight downstream, so arc
length and span decouple completely. On a 16.3 m ram-air kite (SK100, 44 panels,
6.3 m max chord) the `march_edges` stations closest to the tip read:

```
station 251 | y=8.132 | d_arc=0.0816  d_span=0.0032  d_x=0.0807 | arc/span=25.87
station 252 | y=8.135 | d_arc=0.0816  d_span=0.0032  d_x=0.0807 | arc/span=25.87
station 255 | y=8.145 | d_arc=0.0816  d_span=0.0032  d_x=0.0807 | arc/span=25.87
```

Identical rows: the march has settled into walking along the tip cap, spending
**25.87 m of arc per metre of span**. That is faithful to the geometry — the min-x
crossing really does race aft as a section closes — but it means equal arc-length
targets are not equal anything in span.

## What it does to the sections and panels

Sections bunch into the cap. With `wingtip_distance = 0.05`:

```
sec 40: y=7.607   sec 43: y=8.113
sec 41: y=7.893   sec 44: y=8.132
sec 42: y=8.069   sec 45: y=8.151
```

Four sections in the outermost 8 cm; sections 40->41 alone span 29 cm.

These are the panels, not just the sections. With `n_panels = 44` and
`n_sections = 45`, `refine!` takes the `length(unrefined_sections) == n_panels + 1`
branch and calls `copy_sections_to_refined!`, so the declared `BILLOWING`
distribution is never reached and the mapping is 1:1 — measured
`max |panel corner - section point| = 0.000e+00`. Per panel:

```
panel |  y_1     y_2   | d_span
    1 |  8.121   8.110 | 0.0110      <- 1.1 cm
   22 |    ...         | 0.5440      <- 54.4 cm
   44 | -8.110  -8.121 | 0.0110

widest/narrowest = 49.5
```

The 1.1 cm panels are geometrically degenerate. Measured on the undeformed CAD wing
in a uniform freestream:

```
panel |  sweep from freestream | x_airf . y_airf
    1 |               10.7 deg |           0.996
    2 |               24.5 deg |           0.935
   21 |               88.5 deg |           0.028
   44 |               10.7 deg |           0.996
```

The outermost panel's bound vortex sits **10.7 deg from the freestream** and its
chordwise and spanwise axes are nearly parallel. 6 of 44 panels are below 45 deg.

## Why it matters beyond those panels

`q_dyn = 0.5*rho*|v x y_airf|^2` correctly gives such a panel `sin^2(10.7) = 3.4%`
of freestream, so it carries almost no load — that part is fine and should not be
"fixed" by orthogonalising `y_airf`, which would fabricate a factor of 29 in tip
force. The problem is conditioning. `z_airf = normalize(x_airf x span_vec)` is the
normalised cross product of two vectors 5 deg apart, so on this panel **1 cm of
corner motion swings `z_airf` by 33 deg and its angle of attack by 2.44 deg**
(mid-span: 0.05 deg). Since the LOOP iteration solves

```julia
gamma_new[i] = 0.5 * v_a_dist[i]^2 / va_magw[i] * cl(alpha_i) * chord[i]
```

a panel with negligible force still injects a wildly swinging circulation into every
other panel's induced velocity. On a flexible wing this shows up downstream as all
panels going non-finite at once.

## The existing workaround, and why it is a poor one

`wingtip_distance` does help, but being in arc length it inherits the same problem.
Measured panel-level through `obj_to_yaml` on the same mesh:

```
 wtd  | worst sweep | x.y   | panels<45 | span ratio | half-span lost
 0.05 |    10.7 deg | 0.996 |         6 |       49.5 |  0.0%
 1.0  |    41.7 deg | 0.780 |         2 |        2.7 |  1.7%
 1.4  |    66.0 deg | 0.449 |         0 |        2.7 |  4.2%
 1.8  |    69.5 deg | 0.397 |         0 |        2.7 |  9.1%
 2.0  |    70.7 deg | 0.377 |         0 |        2.5 |  9.1%
 2.4  |    72.8 deg | 0.343 |         0 |        2.5 | 14.2%
```

It is strongly non-linear (0 -> 1.2 is nearly inert, because it is eating cap arc)
and it has plateaus (1.8 and 2.0 snap to the same march station). A user asking for
"1.2 m of tip inset" gets 4.9 cm of span.

## Suggested fix

Place the targets by spanwise position rather than leading-edge arc length — either
`march.le[i][2]` directly, or the arc length of the quarter-chord line projected
into the span direction, which is also what `refine_mesh_for_linear_cosine_distribution!`
already uses for `LINEAR`/`COSINE` (`src/wing_geometry.jl:1347`). That makes the
section distribution mean the same thing on a raked tip as on a rectangular wing,
and makes `wingtip_distance` a predictable control.

Smaller and independent: `station_indices` takes `min_chord_frac` (default `0.01`,
i.e. 1% of max chord = 6 cm here, which is why a 0.5 m-chord section survives), but
`perpendicular_sections` calls it as `station_indices(m, n_sections; wingtip_distance)`
and never passes it through, so it is unreachable from `WingSettings`. Plumbing it
would give a direct, geometrically meaningful way to drop degenerate tip stations.

## Version

`main` @ v4.3.1 with #273 applied (quarter-chord normal), `src/obj_adapter/obj_slice.jl`.
Related: #272.

1-Bort-1:
### TL;DR

`station_indices` now places the `n_sections` targets evenly along the quarter-chord line with its chordwise (`x`) step dropped, instead of along the leading edge's arc length. At a raked or closing tip the leading edge runs almost straight aft, so arc-length targets piled into the last few centimetres of span, and `wingtip_distance` gave almost no inset there.

### What was wrong

`march_edges` summed the full 3D leading-edge step into `arclen`, and `station_indices` spread its targets over that. A station where the leading edge moves 40 mm aft for 3 mm of span counts as 40 mm, so the tip cap takes a share of the sections far beyond its span.

The SK100 mesh from the issue isn't in the repo, so I reproduced the problem on a synthetic march: a straight leading edge over |y| ≤ 1 m, then 20 cap stations per side running 40 mm aft per 3 mm of span (half-span 1.06 m). With 9 sections, main gives:

```
y    = [-1.06, -1.027, -0.9, -0.45, 0.0, 0.45, 0.9, 1.027, 1.06]
diff = [0.033, 0.127, 0.45, 0.45, 0.45, 0.45, 0.127, 0.033]   (even spacing: 0.265)
wingtip_distance = 0.3 → outermost y = ±1.039                 (a 0.3 m inset: ±0.76)
```

A 0.3 m `wingtip_distance` moved the outermost section 2 cm. That is the "1.2 m of tip inset gives 4.9 cm of span" from the issue.

### What changed

- `station_indices` computes the quarter-chord point of every marched station and adds up `hypot(Δy, Δz)` between neighbours. The targets, the `min_chord_frac` trim and `wingtip_distance` all use that length now.
- `march_edges` no longer returns `arclen`. `station_indices` was its only reader.
- Docstrings, `docs/src/airfoil_pipeline.md`, `docs/src/settings.md` and the `WingSettings` field doc now describe `wingtip_distance` as a spanwise length [m]. The changelog entry is under Changed.

**Why not plain `y`, the issue's first suggestion?** The repo's ram-air kite curls down at the tips. Near station 80 the leading edge moves `dy=0.0185, dz=-0.1206` per step. Placing sections by `y` would spread the outer ones far apart along the real span. Dropping only `x` handles both a curved span and a raked tip.

**Why the quarter chord?** It is the line the bound vortex lies on, and it is what `refine_mesh_for_linear_cosine_distribution!` already measures. Dropping `x` removes the part of each step that points along the freestream, which is what made those panels degenerate.

### Effect on the repo's ram-air kite

Ram-air kite front view, 10 sections: before, placed by leading-edge arc length; after, evenly over the span

The "before" panel ran main's `obj_slice.jl` (loaded from `origin/main`) on the same march, with the same `n_sections=10`, `wingtip_distance=0` and axes. The largest panel span divided by the smallest drops from 1.44 to 1.10. This mesh has no long raked cap, so the change is modest here. The tips that close to a point are still trimmed as before (tip chord 0.168 m in both).

### Where I'd push back

- **Sections still snap to the nearest marched station.** When `n_sections` gets close to the number of stations, rounding dominates. On this mesh, `n_sections=45` with the default `n_bins=60` gets worse: the span ratio goes from 2.07 to 4.45, because the curled tip stations are about 0.12 m apart in `y`-`z` while the inboard ones are 0.055 m apart. With `n_bins=240` it improves instead, from 1.90 to 1.36. Placing targets finer than the station spacing would mean interpolating between stations, which is a separate change.
- **I couldn't check the SK100 cap.** Its numbers in the issue (`d_arc=0.0816, dx=0.0807, dy=0.0032`) leave about 12 mm of `z` per station. If that `z` is the nose sliding vertically as the section closes, rather than real span, those stations still count about 3.8× their `y` step (it was 25.9×). Someone with that mesh should rerun the panel-sweep table.
- **Old outputs don't regenerate on their own.** The `wingtip_distance` values and generated geometries already out there were placed the old way. `obj_to_yaml` reuses an existing `geometry.yaml`, and `test/generated/` is keyed on settings rather than on the code, so both have to be regenerated.

### Left out

- **Passing `min_chord_frac` through to `WingSettings`.** The issue calls it independent, and it would add a new setting. With a spanwise `wingtip_distance` it may no longer be needed; I'd open it separately if it still is.
- **The `z_airf` conditioning on near-streamwise panels.** That is #272 / #273.

### Verification

- [x] Reproduced first on a synthetic raked-tip march: main gives the spacing and inset shown above. On main the new test's march errors with `FieldError: type NamedTuple has no field arclen`; with `arclen` supplied it fails on the numbers.
- [x] `test/obj_adapter/test_obj_adapter.jl` "station_indices spreads sections evenly in span past a raked tip": green after the change (3/3, juliaserver)
- [x] Affected files, with `test/generated/` rebuilt by this code: `obj_adapter/test_obj_adapter.jl` 55/55 · `settings/test_settings.jl` 33/33 · `test_refinement_validation.jl` 15/15 · `wake/test_wake.jl` 3/3 · `body_aerodynamics/test_results.jl` 30/30 · `solver/test_forwarddiff.jl` 7/7
- [x] `plotting/test_plotting.jl` (calls `station_indices` through `plot_slices_3d`): 58/58 in the local CI mirror
- [x] Docs build clean (`docs/make.jl`) · up to date with main · no REUSE lint in this repo
- [x] Local CI mirror, full `Pkg.test()` on Julia 1.12.7: PASS, 6421 pass and 1 broken (the existing `@test_skip` in `test_kite_geometry.jl`), 14 min · GitHub CI: runs when this PR opens
- Benchmark: n/a
- Risk: generated geometries change section positions. No test pins those numbers, but downstream users with a tuned `wingtip_distance` get a different inset.

### Scope

+56 / −33 across 7 files. `obj_slice.jl` is about net zero: `arclen` comes out of `march_edges` and the spanwise sum goes into `station_indices`. Also in the diff: the new test (+19), the changelog (+4), and docstring and doc wording. Rewrapping the `obj_to_yaml` docstring paragraph is the only tidy-up. I searched for `arclen`, `station_indices`, `wingtip_distance`, "arc length" and quarter-chord helpers; none existed to reuse.

Closes #275 · task `VortexStepMethod.jl-275`


src/obj_adapter/obj_slice.jl
```diff
@@ -455,19 +451,24 @@ end
...
 function station_indices(march, n; wingtip_distance=0.0, min_chord_frac=0.01)
     chords = [norm(te .- le) for (le, te) in zip(march.le, march.te)]
     usable = findall(≥(min_chord_frac * maximum(chords)), chords)
-    arclen = march.arclen
-    inner, outer = arclen[first(usable)], arclen[last(usable)]
+    quarter_chord = [le .+ 0.25 .* (te .- le) for (le, te) in zip(march.le, march.te)]
+    span = zeros(length(quarter_chord))
+    for i in 2:length(span)
+        step = quarter_chord[i] .- quarter_chord[i-1]
+        span[i] = span[i-1] + hypot(step[2], step[3])
```
1-Bort-1, line 468: MINOR: The card says this matches what `refine_mesh_for_linear_cosine_distribution!` measures, but that function (and `compute_refined_section_interpolation!`) uses full 3D quarter-chord length with `x` included. The repo now has two different ideas of span position, and refinement still places panels by the length this PR stops using. Either correct the card and say so, or explain why the two may differ.

1-Bort-1 (commented):

**Good**

- The fix matches the card: `station_indices` now spreads targets over the quarter-chord `hypot(Δy, Δz)` length, and the `min_chord_frac` trim and `wingtip_distance` use that same length (obj_slice.jl:464-473).
- Removing `arclen` from `march_edges` is safe: a repo-wide `rg arclen` finds no other reader, and `plot_slices_3d` in the Makie extension only calls `station_indices`.
- The docs are consistent: settings.jl, settings.md, airfoil_pipeline.md, both docstrings and the changelog all call `wingtip_distance` a spanwise length [m], and `rg 'arc length'` finds nothing stale in the slicer path.
- The new test checks both even spacing and the size of the inset; I worked it through by hand: all 61 stations stay above `min_chord_frac`, the targets land within one 0.05 m station of the expected spacing, and the tolerance holds.
- The change stays in scope: the only tidy-up is rewrapping the `obj_to_yaml` docstring, as the card says, and `min_chord_frac` passthrough and #272/#273 are left out on purpose.

**Not good**

- `src/obj_adapter/obj_slice.jl:468` — The card says this matches what `refine_mesh_for_linear_cosine_distribution!` measures, but that function (and `compute_refined_section_interpolation!`) uses full 3D quarter-chord length with `x` included. The repo now has two different ideas of span position, and refinement still places panels by the length this PR stops using. Either correct the card and say so, or explain why the two may differ.
- The test has z = 0 everywhere, so it only checks that `x` is dropped. Changing the code to use plain `y` would still pass, even though the card's reason for including `z` is the curled ram-air tip.
- If closed tip stations step only in `x`, their `span` values equal `span[first(usable)]`. `argmin` then returns the first of the tied indices, which is a closed station, and that silently undoes the `min_chord_frac` trim on the left tip. This is unlikely on real meshes but worth a line in the card.
- The `station_indices` docstring is 6 lines, above the rubric's 1-4. Its last two sentences could be cut down to state only what is left out and what `wingtip_distance` does.

1-Bort-1:

CHANGELOG.md
```diff
@@ -8,6 +8,10 @@
 - The Makie `plot!` methods for a `Panel` or a `BodyAerodynamics` return a
   `Vector{Makie.AbstractPlot}` instead of a `Vector{Any}`; for a `BodyAerodynamics`
   drawn as flat panels it is one flat list rather than a list per panel.
+- `obj_to_yaml` and `perpendicular_sections` spread the sections evenly over the span,
```
1-Bart-1: this is a breaking change

src/obj_adapter/obj_slice.jl
1-Bort-1: Corrected in #350's description: both refinement paths sum the full 3D quarter-chord step, x included, so the two measures differ; left as a push-back bullet rather than changed here.

CHANGELOG.md
1-Bort-1: Marked `BREAKING:` in 1fe72b9.

1-Bort-1:
**CI: `Julia 1.13 - ubuntu-latest - x64 - pull_request` failed** on `1fe72b9a2`.

https://github.com/OpenSourceAWE/VortexStepMethod.jl/actions/runs/35522342153/job/106108474964

1-Bart-1:
> Refinement measures span differently. refine_mesh_for_linear_cosine_distribution! (src/wing_geometry.jl:1327) and compute_refined_section_interpolation! place LINEAR/COSINE panels by the full 3D quarter-chord length, x included, so the repo now holds two ideas of spanwise position: the slicer's, which drops the chordwise step, and refinement's, which does not. Refinement runs on sections a user may have written by hand, where the same raked-tip collapse would bunch panels rather than sections. I left it alone because changing it moves every refined panel on every existing wing, which is a decision for its own thread rather than a line in this diff.

I think the correct way lies in the middle. Along the arch of the wing, so local y/spanwise, but without the chordwise component. Then you get equal-width panels also with sweep.

1-Bort-1:
Done in 40f69c3, allocation fix in 09d4e84: LINEAR/COSINE refinement and the slicer now measure span with the same spanwise_step, the y-z length of each quarter-chord step with x dropped. A wing kinked into 45° sweep now gets panels of equal span, [2, 1.5, 1, 0.5, 0] where it used to get [2, 1.573, 1.146, 0.604, 0].

1-Bort-1:
Tried in 40f69c3, then reverted in f06df0b after we talked it through: only the slicer drops x, because there the LE turns into the side of the wing at no definite point. Refinement keeps the full 3D quarter-chord length, the bound-vortex span of Gaunaa 2026 CP1.

PR #341, opened 2026-09-16
Rename the `*_array` locals to `_dist`, matching the per-panel fields they hold

1-Bort-1:
### TL;DR

The 22 locals and arguments ending in `_array` (per-panel data, some 1D and some P×3) now end in `_dist`, `calc_norm_array!` is now `calc_norm_dist!`, and `find_center_of_pressure` takes `force` and `moment`. "Array" says nothing about the shape, and the struct fields holding the same data already end in `_dist`.

### The naming decision

#147 asked for `vector`/`matrix` (or `_vec`). I went with `_dist` for everything per-panel, whether it is 1D or 2D, for three reasons:

- The top of `src/solver.jl` already sets the rule: "Variables ending in `_dist`: per-panel distributions".
- The fields these locals read from are already `_dist`: `VSMSolution._x_airf_dist`, `_va_dist`, `_chord_dist`, and `LoopResult.va_norm_dist`/`va_unit_dist`.
- `solver.jl` was already passing `cl_dist` and `chord_array` in the same call.

With `_vec`/`_mat`, a local and the field it reads would have different names (`x_airf_mat = solver.sol._x_airf_dist`). The shape is already in the type, and the docstrings give it where it matters.

Only the suffix changes; the rest of each name stays (`panel_width_array` → `panel_width_dist`, not the field's `width_dist`). That keeps every rename one mechanical rule. The two exceptions are `force_array` and `moment_array`: each holds one 3-vector, not a per-panel quantity, so they lose the suffix. No clashes: before renaming I checked that no new name already exists in the same function.

No public API moves. Everything renamed is a local, a positional argument, or the private `calc_norm_array!`. Nothing renamed is exported, a struct field, or a keyword.

### Also fixed in lines this PR already touches

- In `test_body_aerodynamics.jl`, both AIC testsets sized `va_norm_dist`/`va_unit_dist` with `length(coord)`. `coord` is a matrix, so that counts elements, not panels. They now use `length(body_aero.panels)`. The test result can't change, because `calculate_AIC_matrices!` only reads rows `1:P`.
- Three docstring signatures:
  - `update_effective_angle_of_attack!` was documented under the name of a function that no longer exists (`..._if_VSM`) and left out `alpha_corrected`.
  - `calculate_results` left out `reference_point` and its keywords.
  - `calculate_AIC_matrices!` left out `target`.
- The MWEs (`mwes/mwe_01.jl`, `mwe_warntype.jl`) import `calc_norm_array!`, so they follow the rename.

[...]

PR #352, opened 2026-09-16
Add the Gaunaa et al. 2024 spanwise-flow viscous drag correction as an opt-in solver setting

### LINKED ISSUE #332: Add the optional spanwise-flow viscous drag correction (Gaunaa et al. 2024) from the Python solver
The Python solver has an optional correction for the effect of spanwise flow on the viscous force, which this package lacks. It is off by default there.

`Solver(is_with_viscous_drag_correction=True)` calls `BodyAerodynamics.viscous_drag_correction` (`src/VSM/core/BodyAerodynamics.py:985`, applied at `:1281`, ocayon/Vortex-Step-Method@e5c3b49), following Gaunaa et al. 2024, *A correction model for the effect of spanwise flow on the viscous force contribution in BEM and Lifting Line methods*, J. Phys.: Conf. Ser. 2767 022068, doi:10.1088/1742-6596/2767/2/022068. Per panel, with β the angle between the local inflow and the panel plane normal to the span and Re built on the chordwise speed:

- f0 = 0.062 Re^(-1/7)
- ΔCd = f0 (cos(β)^(-5/7) - 1), along the in-plane drag direction
- C_par = f0 tan(β) cos(β)^(-5/7), along the span

Here `rho` is a solver input and `mu` is only used for the Reynolds number (`src/body_aerodynamics.jl:1002`); searched `viscous_drag|gaunaa|spanwise.?flow|crossflow`, no hits beyond the Li/Gaunaa artificial viscosity.

Relevance: sideslip and anhedral give the outer panels of a kite a spanwise inflow component, which is what this corrects. Proposal: a `SolverSettings` flag, default off, applied where the panel forces are assembled so `solve!`, `solve` and `linearize` all see it. Python's implementation prints per-panel debug output and has no test; a port would want one against the paper's equations.

1-Bort-1:
### TL;DR

`is_with_viscous_drag_correction` (default `false`, in `SolverSettings` and `Solver`) gives each panel the viscous drag increment and spanwise force that Gaunaa et al. 2024 (doi:10.1088/1742-6596/2767/2/022068) predict when flow crosses the span. It runs in both force loops, so `solve!`, `solve` and `linearize` all include it. The equations come from the paper rather than the Python code, because the Python version always adds zero force.

### Why this doesn't copy the Python code

In the Python solver (awegroup/Vortex-Step-Method, the renamed `ocayon` repo), `viscous_drag_correction` gets β from `dir_induced_va = cos(α)·y_airf + sin(α)·x_airf`. That vector lies in the airfoil plane, and Python's `z_airf` is the span axis, so `v_par = Umag·dot(dir_induced_va, z_airf)` is always 0. So β = 0, and both ΔCd and C_par are 0 for every panel. The `Umag` it gets is `|v_eff × z_airf|`, which contains no spanwise part either. Its drag direction `z × (z × d)` is `−d`, pointing upstream, so a non-zero increment would have *reduced* drag. So matching Python's numbers would only prove that this port does nothing as well.

### What the port does

- `spanwise_flow_drag(v_a, v_span, chord, density, mu)` in `panel_aerodynamics.jl` computes the paper's equations: f0 = 0.062·Re^(-1/7), ΔCd = f0·(cos β^(-5/7) − 1), C_par = f0·tan β·cos β^(-5/7). Re and both coefficients use `v_a`, the speed normal to the span. The loops already use that speed for the polar forces (`lr.v_a_dist = |v_eff × y_airf|`), which matches the paper's choice of reference.
- The spanwise velocity was not stored anywhere. `update_gamma_candidate!` now also writes `lr.v_span_dist = dot(v_eff, y_airf)`, in the same pass that computes `alpha` and `v_a`.
- `panel_loads` takes `c_span` (default 0) and adds that force along `y_airf`. In `calc_forces!` and `calculate_results`, the correction adds ΔCd to `cd_dist` and passes `c_span` to `panel_loads`. The per-panel force, the totals and the moments all include it. The flag is also copied into `Solver(body_aero, settings)` and `make_dual_shadow`, so `linearize` keeps it through ForwardDiff.
- Cleanup in `calculate_results`, no change in behaviour: `lift_induced_va` and `drag_induced_va` only ever appeared summed. The loop now projects `loads.force` directly and builds the moment the same way `calc_forces!` does. That is why the file got 24 lines shorter while gaining the feature. `test_body_aerodynamics.jl` still matches the Python reference results (4870 tests).

Size of the effect, on the 8 m × 1 m INVISCID test wing at 20 m/s and α = 6°:

| β | side force added | drag added |
|---|---|---|
| 0° | 0 | 0 |
| 12° | 3.35 N | 0.94 N |
| 30° | 7.92 N | 5.12 N |

Even with no sideslip, the chordwise trailing segments induce a small spanwise velocity at the tip panels, so the correction there is up to 2.4 mN per panel. It cancels in the total.

### Where I'd push back

- Not stacked on #340, although both edit `src/solver.jl`. #340's head predates #341, which this change builds on, and stacking copied #341's commits into this branch. The only real overlap is one line: the new flag in `Solver(body_aero, settings)`. Whichever PR merges second moves that line into #340's `solver_kwargs`.
- SymbolicAWEModels builds its equations from `panel_loads` with symbolic arguments. The new `c_span = 0` term should simplify away there, but I haven't run SymbolicAWEModels against this branch.
- Searched for `viscous_drag|gaunaa|spanwise.?flow|crossflow|cross_flow|v_span`: the only match is the Li/Gaunaa artificial viscosity, nothing to reuse.

### Verification

- [x] Reproduced first: n/a, new feature. Python's zero result comes from reading its code (the axes are orthogonal), not from running it.
- [x] `test/solver/test_viscous_drag_correction.jl`: before the change it errored (`UndefVarError: spanwise_flow_drag`); now 90/90 pass in a fresh juliaserver session. It checks `spanwise_flow_drag` against the paper at β = −40…35°, and checks each panel's drag and spanwise force for sideslip and straight inflow. It also checks that the drag rises along the flow, that `solve` and `linearize` report the corrected forces, that `calc_forces!` stays zero-alloc with the flag on, and that the flag defaults to off.
- [x] Other affected files pass in a fresh session: `test_solver` 34, `test_unrefined_dist` 40, `test_body_aerodynamics` 4870, `test_results` 30, `test_flow_curvature` 39, `test_moment_units` 17, `test_backend_comparison` 15, `test_forwarddiff` 7, `test_verification` 22, `test_settings` 33, `test_panel` 26
- [x] Docs build clean (only the existing page-size warnings) · no REUSE in this repo · up to date with main
- [x] Local CI mirror (`agent ci-local`, Julia 1.12.7): PASS (11 min, exit 0) · GitHub CI: PASS
- [x] GitHub CI's first run failed one check: `Julia 1.12 - ubuntu` at `test_forwarddiff.jl:87` (POLAR_MATRICES), `relative_error` 0.0460. This is #287, not this branch. Re-running the job unchanged made it green. The three test cells that passed print Jacobian norms identical to all 16 digits to `main` @ 246237a on both lookup tables. Numbers are on #287.
- Benchmark: n/a. The flag defaults to off, and with it off `calc_forces!` stays zero-alloc (`test_solver.jl`, and the new test with it on).
- Risk: the paper's coefficients assume a flat-plate turbulent boundary layer (Re^(-1/7)); a user who turns this on for a low-Re or laminar section gets that model anyway.

### Scope

+185 / −54 across 8 files: 100 lines of new test; `calculate_results` −24 net after the loop cleanup; `spanwise_flow_drag` and the `c_span` term +20; the flag, `v_span_dist` and its wiring +26 across `solver.jl` and `settings.jl`; CHANGELOG, docs index and `runtests.jl` +9.

Closes #332 · task `VortexStepMethod.jl-332`


1-Bort-1:
**CI: `Julia 1.12 - ubuntu-latest - x64 - pull_request` failed** on `4965a5fad`.

https://github.com/OpenSourceAWE/VortexStepMethod.jl/actions/runs/35159906923/job/105008016132

1-Bort-1:
**Local full suite: PASS** (9 min, Julia 1.12.7, one cell of the matrix)

1-Bart-1 (approved):

Issue #113, opened 2025-03-09
Add new features from Python version


ufechner7:
see: https://github.com/ocayon/Vortex-Step-Method/blob/develop/changelog.md

#### Done
- Moment Calculations: Added moment calculations and plotting functionality (# 89).  
  Not clear: which plotting functionality?
- User Configuration: Enabled support for a user-defined reference point in moment calculations.
- Renamed WingAerodynamics to BodyAerodynamics.

#### Planned
- NeuralFoil Integration: Integrated NeuralFoil to verify and improve polar predictions. #105 
- Introduced a from_file class method for easier input handling ([# 115](https://github.com/ocayon/Vortex-Step-Method/issues/115)). --> wait until Jelle says the data format is ready, see #133 
-  Bridle Line Forces: Added aerodynamic force functions for bridle lines (# 105), supporting input in the form:  
   bridle_line1 = [p1, p2, diameter] where p1 = [x, y, z]: 
   bridle_lines = [bridle_line1, bridle_line2, ...]
   Could be implemented in a type LEI_Kite, should be implemented by Jelle or students
- Panel Polar Investigation: Implemented a panel polar investigation function (# 92).  
  Should be a plotting function that plots the polars of a panel --> Jelle

#### Not planned
- Plot Enhancements: Enhanced the interactive plot with additional parameters (span, height, chord) and decimal support.
- Interactive Visualization: Introduced an interactive Plotly environment for data visualization.

ufechner7:
@jellepoland Can you comment on this list, in particular the last two points?

1-Bart-1:
If no fancy method is used for bridle line forces in the python package, I would prefer to keep it in KiteModels.jl

jellepoland:
The panel polar investigation refers to a set of plotting functions that can be used to make 2D polar plots. Cl, Cd, Cm alphas, to investigate how the used .csv / polynomial/inviscid / or other underlying method looks like. Is it smooth? Yes/no, is the sign what you expect, etc? 

The "instantiate_body_aerodynamics" has become a class method.
Which is to make the user's life easier when working with a geometry defined as a CSV file.

```
@classmethod
    def from_file(
        cls,
        wing_instance,
        file_path,
        is_with_corrected_polar=False,
        path_polar_data_dir="",
        is_with_bridles=False,
        path_bridle_data=None,
    ):
        """Instantiate a BodyAerodynamics object with or without bridles

        Args:
[...]

PR #354, opened 2026-09-16
plot_section_polars draws cl, cd and cm against alpha per panel, for every aero model

### LINKED ISSUE #331: plot_section_polars covers only POLAR_VECTORS: plot cl, cd, cm against alpha per panel for every aero model
The "panel polar investigation" planned in #113 is only partly there: `plot_section_polars` plots one of `:cl`, `:cd`, `:cm` per call and only for sections with `POLAR_VECTORS` data (`ext/VortexStepMethodMakieExt.jl:1508`; other sections are skipped, and it errors when none are left). `plot_polar_data` draws `POLAR_MATRICES` as 3D (alpha, delta) surfaces for panel 1 only. Nothing plots what a panel actually uses for `POLY` (Breukels), `INVISCID` or `POLAR_MATRICES` at a given flap deflection.

What @jellepoland described in #113: 2D plots of cl, cd and cm against alpha, to check whether the underlying method — CSV polar, polynomial, inviscid or anything else — is smooth and has the expected sign. Python had it as `plot_panel_coefficients(body_aero, panel_index, alpha_range=[-20, 30])` (ocayon/Vortex-Step-Method@48d4e93, "#92 panel polar investigation"); it is commented out on Python main today (`src/VSM/plotting.py:700`), so there is nothing current to port.

Proposal: one figure per call with cl, cd and cm against alpha, evaluated through the panel's own coefficient functions (`calculate_cl`, `calculate_cd`, `calculate_cm` in `src/panel.jl`) so every aero model goes through the same path, for a chosen panel or list of panels, with the flap deflection as a keyword for `POLAR_MATRICES`. This extends `plot_section_polars` rather than adding a second function.

1-Bort-1:
### TL;DR

`plot_section_polars(body_aero; panels, alphas, delta)` now draws cl, cd and cm in one figure, one curve per chosen panel, through the panel's own `calculate_cl`/`calculate_cd`/`calculate_cm`. Before, it drew one coefficient per call from raw `POLAR_VECTORS` section data only. A `POLY`, `INVISCID` or `POLAR_MATRICES` wing gave `No POLAR_VECTORS sections found in body`, and nothing showed what a panel uses at a given flap deflection (#113).

### What changed

- **One path for every model.** A new helper, `panel_polar_curves(panels, alphas, deltas)`, runs the same `calculate_*` functions the solver calls over an α grid. The plot therefore shows what the solver sees: the averaged, interpolated panel polar, with its extrapolation and `alpha_window`. It no longer shows the unrefined section's CSV points.
- **Flap deflection.** `delta` [rad] sets the deflection a `POLAR_MATRICES` panel is evaluated at. It defaults to each panel's stored `delta`, and other models ignore it.
- **Keywords and figure.**
  - `panels` takes one index or a list and defaults to all panels.
  - `alphas` is in radians, like `plot_polar_data`'s. The default −20°…30° is the range of the Python `plot_panel_coefficients`.
  - The figure is one `MakieControlPlots.plotx` with three rows on a shared α axis.
- **Changelog.** Listed under Added. The positional `coefficient` argument is gone, so `plot_section_polars(body_aero, :cd)` now raises a `MethodError`; its one caller, `examples/obj_to_yaml_kite.jl`, now makes a single call.
- **Cleanup in the same file, near-identical behaviour.** The `POLAR_VECTORS` panel of `plot_combined_analysis` used to call the raw `cl_interp`/`cd_interp`/`cm_interp` for the same three curves; it now uses the same helper. The two paths differ only where `alpha_window > 0`, which the raw interpolants skipped. On `create_body_aero_with_skin` over −5…20° the old and new values are identical (`alpha_window = 0`).

after: plot_section_polars on a POLY (Breukels) panel, lei_poly_coeffs(0.1, 0.08); cl and cd jump at alpha = 20 deg

after: plot_section_polars on the TU Delft V3 kite, POLAR_VECTORS panels 1 and 25

There is no before figure. Old code raised an error on the POLY wing, and on the V3 kite it plotted different data: raw section points, not panel polars.

### Found on the way (not changed here)

- **`POLY` is discontinuous at |α| = 20°**, which the plot shows at once: at `lei_poly_coeffs(0.1, 0.08)` cl drops from 1.0706 to 0.2200 across 20°, and cm's quadratic reaches −5.53 at 30°. Fixing it changes existing `POLY` results, so it is #359.
- **`plot_combined_analysis` draws its `POLAR_MATRICES` wireframes transposed.**
  - The matrix is built `for alpha in alphas, delta_te in delta_tes`, which puts α on the rows. It is then drawn with `wireframe!(ax, delta_tes, alphas, M)`, which expects δ on the rows.
  - `plot_polar_data` builds the matrix the other way round and is correct.
  - Both grids are `-5:0.3:20`, so nothing errors. I'd fix it by having both functions share one wireframe helper; that changes a figure, so it is not in this diff.
- **CI matrix.** `.github/workflows/CI.yml:22` carries `fail-fast: false`, so one failing cell does not cancel the others and a red run costs the whole matrix; turning it off is a `cleanup:` PR of its own.

### Verification

- [x] Reproduced first, on unchanged code: `plot_section_polars(create_body_aero(), :cl)` (INVISCID) and `plot_section_polars(BodyAerodynamics([ram_wing]), :cl)` (POLAR_MATRICES) both raise `No POLAR_VECTORS sections found in body`.
- [x] The new testset in `test/plotting/test_plotting.jl` was red before the change (`MethodError: no method matching plot_section_polars(...; panels, alphas)`) and is green after, 19/19. It has one nested testset per behaviour:
  - INVISCID gives cl = 2πα and cd = cm = 0;
  - POLAR_VECTORS gives cl = 0.5 + 0.25·α°, cd = 0.02 and cm = −0.05;
  - POLY matches `evalpoly` of the Breukels coefficients;
  - POLAR_MATRICES at `delta = 1°` matches the interpolants there and differs from the stored `delta`;
  - panel selection, and the saved file.
- [x] All testsets in `test/plotting/test_plotting.jl` pass in the juliaserver session: Plotting 58/58, the new testset 19/19, Airfoil skin 22/22, generated_slices 6/6, Audit slices 7/7.
- [x] `plot_combined_analysis` on a POLAR_VECTORS body still builds a `Figure` (checked by hand; no test covers that branch).
- [x] Docs build (`docs/make.jl`) is clean apart from the existing size-threshold warnings. The repo has no REUSE.
- [x] Up to date with main: merged `origin/main` twice, most recently at 7103702 (after #349's `va` rename, #351 and #355). The only conflict was the changelog's Added list, where both entries are kept. `test/plotting/test_plotting.jl` was re-run after each merge, and all testsets pass as above, now on Julia 1.13.0.
- [x] GitHub CI on 7103702: PASS on every job (Julia 1.12 ubuntu/windows/macOS, 1.13 ubuntu, Documentation, setup test). Local full suite on 7103702: PASS (13 min, Julia 1.13, one cell).
- Benchmark: n/a
- Risk: scripts calling `plot_section_polars(body_aero, :cd)` break. The old form had no test and one example caller.

### Scope

+126 / −53 across 6 files against main:
- `ext/VortexStepMethodMakieExt.jl`: the function body, now shorter, plus the helper and the three lines in `plot_combined_analysis`.
- The stub docstring and its private-docs entry.
- The testset.
- The example call and the changelog entry under Added.

Searched for `calculate_cl`, `cl_interp(`, `plot_polar`, `panel_coefficients` and `alpha_range` in `src/` and `ext/`. The only other code that evaluates a panel over an α grid is `plot_combined_analysis`, which is folded in above.

Closes #331 · task `VortexStepMethod.jl-331`


1-Bort-1 (commented):

**Good**

- The plot now reaches every aero model: `calculate_cl/cd/cm` in `src/panel.jl:504-580` handle POLY, INVISCID, POLAR_VECTORS and POLAR_MATRICES, and use `delta` only for POLAR_MATRICES, as the docstring says.
- One code path, as the card promises: `panel_polar_curves` replaces the three raw `*_interp` calls in the POLAR_VECTORS branch of `plot_combined_analysis`, so no third copy exists.
- Panels and deflections pair up correctly: `calculate_cl.(panels, alphas', deltas)` with a per-panel `deltas` vector gives an n×m grid, `something.(nothing, …)` works because broadcasting treats `nothing` as a scalar, and `vcat(panels)` accepts both an Int and a range.
- `labels=[labels]` fits `MakieControlPlots.plotx` (`plotx.jl:27-37`): the panel legend goes only on the first (cl) row, which is right for three rows sharing an α axis.
- The tests pin each model to a closed form (2πα, the linear table, `evalpoly` of the Breukels coefficients, the POLAR_MATRICES interpolant at the passed delta) and do not just check that a figure exists; they reuse `create_body_aero` through new keywords, not a new fixture family.
- The BREAKING change is small: a search finds only the one example caller, updated in this diff, plus the changelog and docs stubs; the new helper is listed in `private_functions.md`.
- The scope matches the card: six files, the combined-analysis tidy-up is named, and the found problems (POLY jump at 20°, transposed wireframe) are noted, not fixed silently.

**Not good**

- `ext/VortexStepMethodMakieExt.jl:1521` — By default `panels` is every panel, so a normal 20–40-panel wing draws 20–40 overlapping curves and a legend with that many entries on the cl row; the one example caller already overrides it with `[1, 10]`, which suggests the default is not a useful figure.
- The title is still "Section polars" and the file `section_polars.png`, but the figure now shows panel polars (averaged over the refined panels); the function name is API, but the title string could say what is drawn.
- The two found problems (POLY discontinuity at |α| = 20°, transposed POLAR_MATRICES wireframe in `plot_combined_analysis`) have no issue yet; they are left to the human, where §5 says to open issues and link them from the PR.
- The wireframe transposition is in the same `if` block this PR edits, and fixing it changes only which axis is labelled what; it could arguably have gone into this diff, not just been noted.
- The CI matrix `fail-fast: false` bullet does not say what is wrong with it, so a reader cannot tell why it is listed under found problems.
- Local CI and GitHub CI are both still pending in the card, so the only green result so far is from the juliaserver session.
- The `panel_polar_curves` docstring allows "one shared deflection" only for the `plot_combined_analysis` caller; `plot_section_polars` always passes a vector, so the helper supports two argument shapes for two callers.

ext/VortexStepMethodMakieExt.jl
```diff
@@ -1499,48 +1499,43 @@ function VortexStepMethod.plot_combined_analysis(
...
-            elseif length(section_alphas) != length(alphas_deg)
-                @warn "section $s has a different α grid; plotting against the first section's α"
-            end
-            push!(series, Float64.(aero[idx]))
-            push!(labels, "section $s")
-        end
-    end
-    isempty(series) && error("No POLAR_VECTORS sections found in body")
+function VortexStepMethod.plot_section_polars(body_aero::BodyAerodynamics;
+    panels=eachindex(body_aero.panels), alphas=deg2rad.(-20:0.5:30), delta=nothing,
```
1-Bort-1, line 1499: MINOR: By default `panels` is every panel, so a normal 20–40-panel wing draws 20–40 overlapping curves and a legend with that many entries on the cl row; the one example caller already overrides it with `[1, 10]`, which suggests the default is not a useful figure.

CHANGELOG.md
```diff
@@ -8,6 +8,9 @@
 - The Makie `plot!` methods for a `Panel` or a `BodyAerodynamics` return a
   `Vector{Makie.AbstractPlot}` instead of a `Vector{Any}`; for a `BodyAerodynamics`
   drawn as flat panels it is one flat list rather than a list per panel.
+- BREAKING - `plot_section_polars(body_aero; panels, alphas, delta)` draws cl, cd and cm
```
1-Bart-1: This is not a breaking change but an added feature

1-Bart-1 (changes requested):

CHANGELOG.md
1-Bort-1: Moved to Added in 9ae2ab2. The old positional form, plot_section_polars(body_aero, :cd), now raises a MethodError; that is the one thing a script could trip on.

1-Bort-1:

1-Bart-1 (approved):

PR #355, opened 2026-09-16
linearize and make_dual_shadow take a BodyAerodynamics with several wings

### LINKED ISSUE #344: linearize and make_dual_shadow accept a BodyAerodynamics with more...

1-Bort-1:
### TL;DR

`linearize` and `make_dual_shadow` no longer throw on a body with more than one wing: the dual shadow copies every wing, and `theta_idxs`/`delta_idxs` run over the unrefined sections of all wings in order. #330's stability derivatives need this to cover multi-wing bodies.

### What changed

Only the one-wing guards stood in the way. `solve!` already handled several wings, and `calc_forces!` already filled `moment_unrefined_dist` wing by wing, offset by the section counts of the wings before. The inputs now use that same order, from the same place.

- `make_dual_shadow` builds the shadow body from `_wing_with_eltype` of every wing, not just `wings[1]`. It also passes on #352's `is_with_viscous_drag_correction`, which came in with the merge from main.
- A private `unrefined_section_range(body_aero, wing_idx)` gives the indices of one wing's unrefined sections. `calc_forces!` now uses it instead of its own running offset, so the outputs and the inputs share one definition of the order.
- A private method `unrefined_deform!(body_aero, theta_angles, delta_angles)` hands each wing a view of its own range of angles. `linearize` calls it, then `reinit!` as before.
- `linearize` checks `theta_idxs`/`delta_idxs` against `U`, the solver's total number of unrefined sections, which is also the length of `moment_unrefined_dist`. The result length is `6 + U`.
- Cleanup nearby, no behaviour change:
  - `om` is renamed `omega` and `be` is renamed `ad_backend`.
  - `s` is renamed `section` in `_wing_with_eltype`.
  - The unused `smooth=false` argument is no longer passed.
  - A `wing_pair` test helper replaces four copies of the same two-wing setup.

### Tests

Three testsets under `solve! on a two-wing body` in `test/body_aerodynamics/test_body_aerodynamics.jl`.

- **Each wing gets its own angles.** `unrefined_deform!` on the body with a different twist and deflection for every section must give each wing the same `theta_dist` and `delta_dist` as deforming that wing on its own with its range of angles. The expected ranges are written out by hand, not taken from the helper. The INVISCID Jacobians below cannot see deflection, so this is the only check on the deflection split.
- **Wings 1e4 m apart.** Each wing's twist moves only its own sections' moments. Twisting wing 2 moves wing 2's rows the same way twisting wing 1 moves wing 1's: measured relative difference 1.3e-12, test bound 1e-9. The cross blocks are 5.7e-9 of the diagonal blocks, test bound 1e-7. If the split into wings were wrong, these would be off by O(1).
- **Wings one chord apart**, so they induce on each other. The ForwardDiff Jacobian matches `AutoFiniteDiff` (steps 1e-6) to a relative difference of 2.7e-7. The test allows 1e-4.

### Notes for the reviewer

- **Error message:** a body with no unrefined sections and some `theta_idxs` now gets the length-mismatch `ArgumentError` instead of "Cannot use theta_idxs…". It is the same exception type.
- **Still one wing:** `solve!` and `linearize` on a multi-wing body throw with `type_initial_gamma_distribution=ELLIPTIC`, from the guard in `calculate_circulation_distribution_elliptical_wing`. The default (`ZEROS`) works. Follow-up: #357.
- **Now reachable through `linearize`:** `calc_forces!` and `calculate_results` use `wings[1].spanwise_direction` for the force directions of every panel, so wings with different span directions (a vertical fin, say) get wrong forces from wing 2 onward. The fix changes results, so it is not made here. Follow-up: #358.
- **A wing with no unrefined sections** cannot reach `unrefined_deform!(body_aero, ...)`: `BodyAerodynamics` refuses a wing that has not been refined ("Wing 2 has not been refined"), and refining needs sections.
- **CI setting:** `.github/workflows/CI.yml` still sets `fail-fast: false`. That is a separate `cleanup:` PR, not done here.

### Verification

- [x] Reproduced first: both `linearize` testsets errored with `ArgumentError: Linearization only works for a body_aero with one wing` (12 passed, 2 errored), on the first version of those tests.
- [x] Deflection split, red then green: with the body method handing every wing `delta_angles[1:3]`, `test_body_aerodynamics.jl` gives 4884 passed, 1 failed (the new testset, `delta_dist` of wing 2). All other tests, the Jacobian tests included, still passed. Restored: 4885/4885.
- [x] At 574b15b (juliaserver): `test_body_aerodynamics.jl` 4885, `test/solver/test_unrefined_dist.jl` 40, `test/solver/test_forwarddiff.jl` 7, `test/solver/test_viscous_drag_correction.jl` 90 (includes `calc_forces!` zero-alloc), `test/body_aerodynamics/test_results.jl` 30, `test/solver/test_solver.jl` 34. No failures.
- [x] Merged main twice. The first merge brought in 1c493cf (#352), where only `CHANGELOG.md` conflicted and both entries are kept. The second brought in 83867a3 (#349's va rename, #328), where one set of lines in `linearize` conflicted: main renamed `va` to `va_vec` and this branch renamed `om` to `omega`, and both renames are kept.
- [x] At merge 0b28330, on a relaunched session: `test_body_aerodynamics.jl` 4885, `test/solver/test_forwarddiff.jl` 7, `test/body_aerodynamics/test_results.jl` 30, `test/solver/test_unrefined_dist.jl` 40, `test/solver/test_viscous_drag_correction.jl` 90, `test/solver/test_solver.jl` 41. No failures.
- [x] Local CI mirror on 0b28330: PASS (9 min, Julia 1.12.7) · GitHub CI on 0b28330: PASS, all 7 checks. On the earlier head 574b15b, `Julia 1.12 - ubuntu-latest` failed only #287's POLAR_MATRICES check (0.0460232000169766, table A), and this PR's tests passed in that job.
- Earlier Windows flake, not this change: the first Windows 1.12 run on 1d76212 failed #287's POLAR_MATRICES check at its known value and, in the same process, this PR's FD comparison. A re-run on the same commit passed both. The 16 Jacobian entries in that log are bit-identical to a local run, where the relative error is 2.7e-7. The numbers are on #287.
- Docs build: not run locally, because the docs environment has no manifest here; the CI `Documentation` job passes on 0b28330. `unrefined_section_range` is listed in `docs/src/private_functions.md`. No REUSE in this repo.
- Benchmark: n/a
- Risk: a body whose wings have different panel counts gives the shadow a non-concrete `Vector{Wing{P,TD} where P}`, as the Float64 body already has. It works but dispatches dynamically, and no test covers mixed panel counts.

### Scope

+135 / −47 across 5 files. `src/solver.jl` shrinks (−13 net): three per-wing checks became one loop, and `calc_forces!` lost its running offset. `src/body_aerodynamics.jl` +31 is the range helper and the body method. The test file's +67 net is the three testsets and the `wing_pair`/`linearize_body` helpers.

Closes #344 · task `VortexStepMethod.jl-344`


test/body_aerodynamics/test_body_aerodynamics.jl
```diff
@@ -512,6 +512,32 @@ function solve_wings(wings)
...
+
+`linearize` of `body_aero` at zero twist and deflection over the twist and the deflection of
+every unrefined section, then the inflow and the angular rate.
+"""
+function linearize_body(body_aero; kwargs...)
+    n_sections = sum(wing -> wing.n_unrefined_sections, body_aero.wings)
+    solver = Solver(body_aero; use_gamma_prev=false, rtol=1e-10)
+    y0 = [zeros(2n_sections); body_aero.va; zeros(3)]
+    return VortexStepMethod.linearize(solver, body_aero, y0;
+        theta_idxs=1:n_sections, delta_idxs=n_sections+1:2n_sections,
```
1-Bort-1, line 538: MINOR: `delta_idxs` is split per wing, but the INVISCID wings ignore deflection, so every delta column is zero and no assertion reads them. A swapped or shifted delta offset in `unrefined_deform!(body_aero, ...)` would pass both testsets.

1-Bort-1 (commented):

**Good**

- Matches the card: the one-wing guards in `make_dual_shadow` and `linearize` are removed, and the only new source is `unrefined_deform!(body_aero, ...)`; `git diff 246237a --stat` shows +103/−41 across the 4 files the card names
- `U` really is the total number of unrefined sections across wings (`src/solver.jl:194` sums `n_unrefined_sections`), so `n_results = 6 + U` gives the same size the old `length(moment_unrefined_dist)` did
- Angles are split per wing in the same order `calc_forces!` fills `moment_unrefined_dist` (the `unrefined_idx += wing.n_unrefined_sections` loop at `src/solver.jl:474-520`), so inputs and outputs line up
- Three validation branches became one loop with no change in behaviour except the error message the card mentions; no test checks the old 'Cannot use theta_idxs' text (grep finds none)
- Dropping `smooth=false` changes nothing: it is already the default in `unrefined_deform!(wing, ...; smooth=false)` at `src/wing_geometry.jl:428`
- The far-apart testset would catch a wrong offset: if wing 2 were handed wing 1's slice, the second diagonal block would be zero or off by O(1) against `own_first`
- The renames (`om`→`omega`, `be`→`ad_backend`, `s`→`section`) and the `wing_pair` helper are named in the card as cleanup and are plain Julia locals and test code, not API
- The new docstring is reachable: the bare `unrefined_deform!` entry in `docs/src/private_functions.md:84` picks up every method's docstring
- Every added line fits in 92 columns; the long lines in `src/body_aerodynamics.jl` were already there

**Not good**

- `test/body_aerodynamics/test_body_aerodynamics.jl:536` — `delta_idxs` is split per wing, but the INVISCID wings ignore deflection, so every delta column is zero and no assertion reads them. A swapped or shifted delta offset in `unrefined_deform!(body_aero, ...)` would pass both testsets.
- The far-apart test uses `rtol=1e-4` and a `1e-4` cross-block bound, but the card measured 1.3e-12 and 5.7e-9, so the test allows about 8 orders of magnitude more than it needs to pin the split
- `unrefined_deform!(body_aero, ...)` is a third place that works out per-wing unrefined offsets (with `calc_forces!` and `calculate_results`); a shared per-wing range helper would give one source for that order
- If one wing has no `non_deformed_sections`, the per-wing call now throws 'Call refine!' even for an empty view, where the old code gave a clear idxs error; this is an edge case and nothing tests it
- The shadow body built from wings with different panel counts dispatches dynamically, and no test covers it; the card flags this and it is not blocking
- The card lists the ELLIPTIC guard and `wings[1].spanwise_direction` in `calc_forces!` as known limits but links no follow-up issue; now that `linearize` reaches the second one, it should have an issue

1-Bort-1:
**CI: `Julia 1.12 - windows-latest - x64 - pull_request` failed** on `1d76212db`.

https://github.com/OpenSourceAWE/VortexStepMethod.jl/actions/runs/35161787519/job/105013970853

1-Bort-1:
**Local full suite: PASS** (8 min, Julia 1.12.7, one cell of the matrix)

1-Bart-1:
Address the comments, fix the conflicts, and open a new issue to allow multi-wing to work with elliptic distribution.

test/body_aerodynamics/test_body_aerodynamics.jl
1-Bort-1: Fixed in 574b15b: the new testset "unrefined_deform! hands each wing its own run of angles" fails when wing 2 is handed wing 1's deflections (4884 passed, 1 failed).

1-Bort-1:
Conflicts: merged main in 4ccdea6 and again in 0b28330. In the second merge, `linearize` keeps main's `va_vec` and this branch's `omega`. Review comments: 574b15b (deflection-split test, far-apart bounds tightened to 1e-9/1e-7, one `unrefined_section_range` for `calc_forces!` and `unrefined_deform!`). ELLIPTIC on several wings: #357. `wings[1].spanwise_direction`: #358.

1-Bart-1 (approved):

